PR #84, opened 2020-08-13
[DS-388] pipeline sound zones

cvriel:
Remarks:

1) Source contains two files of which metro, spoorwegen and industrie are located together in one file. Schiphol is in its own separate file.
2) In addition, themas data is used to compliment the thema dimension context to metro, spoorwegen, industrie en Schiphol.
3) Due to .csv import, some extra steps where needed to get geometry as a geometry datatype
4) Check where made on geometry as part of process

Resolves: [DS-388]

[DS-388]: https://datapunt.atlassian.net/browse/DS-388

### DIFF
--- a/src/dags/geluidzones.py
+++ b/src/dags/geluidzones.py
@@ -1,0 +1,270 @@
+import operator
+from airflow import DAG
+from airflow.models import Variable
+from airflow.operators.bash_operator import BashOperator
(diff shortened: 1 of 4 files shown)


cvriel:
> Is een aardige lap code.
> Het ziet er goed uit wat mij betreft.

Ja klopt, had wat meer voeten in de aarde dan een 'normale' dag

PR #90, opened 2020-08-20
Added temp table specification to provenance rename operator

cvriel:
Currently, the provenance operator works with table names which are identical in the schema specification. For temporary tables, which are at the end of the data processing renamed to its final name, cannot be used in the provenance rename operator. The workarround was to keep temp table names the same als the table.id in the schema specification. But that raises problems when different datasets have the same table names. This change will support the specification of the temp table name in the standard manner of {dag_id}_{table_name_as_specified_in_schema}_new

### DIFF
--- a/src/plugins/provenance_rename_operator.py
+++ b/src/plugins/provenance_rename_operator.py
@@ -18,8 +18,8 @@ class ProvenanceRenameOperator(BaseOperator):
         pg_schema="public",
         postgres_conn_id="postgres_default",
         rename_indexes=False,
-        prefix_table_name=None,
-        postfix_table_name=None,
+        prefix_table_name="",
+        postfix_table_name="",
         *args,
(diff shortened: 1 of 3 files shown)


jjmurre (approved):
Lijkt me prima zo! Ik heb nog 1 mogelijke verbetering in een commentaar gezet.

src/plugins/provenance_rename_operator.py
```diff
@@ -17,6 +18,8 @@ def __init__(
         pg_schema="public",
         postgres_conn_id="postgres_default",
         rename_indexes=False,
+        prefix_table_name=None,
```
jjmurre, line 21: Je zou deze prefixes ook de initiële waarde "" kunnen geven. Dan heb je de if-else contstructies op regel 34 en 35 niet nodig.

PR #127, opened 2020-10-08
[DS-464] change for proces two new fields in Horeca

cvriel:
The two new colomns are abbreviated by the source limitation on the column length. In the change, the columns are renamed to more understandable naming.


jjmurre (approved):
Betere namen idd.

PR #140, opened 2020-11-05
[DS-486] removed drop schema because of collision

cvriel:
Due to the use of the same schema (PTE) in multiple dags (source files), the drop schema is removed from the logic to avoid collision.

The drop schema was introduced to cope with the situation that the schema definition had not (yet) specified a table that was already given in the source files. Due to the database dump import the unspecified tables remained in the schema and where not deleted. The deletion is base on the schema specification.

In order to avoid an error between schema definition and source files, it is possible to list the tables that are not yet specified in the schema to be removed after each run.


jjmurre (approved):
LGTM

PR #148, opened 2020-11-20
Remove unwanted LTRIM for meldingen_wijk_id

jjmurre:
The LTRIM was needed for relation to bag:wijk
Now we relate to gebieden:wijken, the 0-prefix does need to stay in the wijk_id


vandoornik:
Mooie fix

vandoornik (approved):

PR #155, opened 2020-12-02
[COVID19] removed check on gebiedsverbod

cvriel:
Since 01-12-2020 the "gebiedsverbod" area's are absolete. They are therefor not checked in Airflow for records. The table still exists so the DSO-API (although empty results) and Schema definition still apply until further notice


jjmurre (approved):
Prima.

PR #156, opened 2020-12-02
[WIP] log cleaner Airflow

cvriel:
- A separate DAG for cleaning logs, can be scheduled to run
- Takes an environment variable to setup the retention time in days, currently set to 90
- Takes an environment variable to indicate the deletions of child folders 
- Looks at the log path as described in the base_log_folder variable in airflow.cfg file


ngaranko (approved):

bheupers (approved):
Lijkt me prima

PR #162, opened 2020-12-16
[GREX] changed startime and convert geometrycollections

cvriel:
- Starttime is change from default to 0600 a.m.
- Source contains geometrycollections (polygons and points in one object)

The startime has changed to a line with the data processing schedule of
DWH stadsdelen. Which provide us with GREX data.

Due to geometrycollections (probably due to source errors) some records could not be processed correctly.
The effect was that these records where not processed.
In order to show the underlining polygons, the additional points in the object are removed.
The order of the convertion steps were not correctly defined.

### DIFF
--- a/src/dags/grex.py
+++ b/src/dags/grex.py
@@ -35,9 +35,8 @@ def wkt_loads_wrapped(data):
     elif isinstance(p, MultiPolygon):
         pass
     else:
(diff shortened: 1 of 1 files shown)


jjmurre (approved):
Lijkt me prima zo.

PR #163, opened 2020-12-17
Make GOB graphql endpoints configurable

jjmurre:
GOB wants to change these endpoints and there will be a considerable
timegap between changing ACC and PRD. By making the endpoints
configurable with an env. var, the Openstack ansible configuration
can be used to manage this change-process.


cvriel (approved):
Ziet er goed uit

PR #164, opened 2020-12-21
[update overlastgebieden] temporary no content check on VVZ

cvriel:
Due to the national meassures of covid19 the "Vuurwerkvrijezones" data is not provided by the source maintainer.
Therefor the check on number of rows and valid geometry is disable for VVZ.

Without the disable, Airflow will raise an error stating the data is not correct.


jjmurre (approved):
LGTM

PR #165, opened 2020-12-22
[FIX] evenementen count check

cvriel:
The minimal number of rows for evenementen data was adjusted from 100 to 75.

The data source maintains lower number of evenementen.
The check for the minimal number of rows was causing Airflow to raise an Error.


jjmurre (approved):
LGTM

PR #166, opened 2020-12-23
[FIX] basiskaart damwand and geluidscherm not in BGTPLUS

cvriel:
The creation of the table SELECT_INRICHTINGSELEMENTLIJN_SQL contains
a selection on damwand and geluidscherm from the BGTPLUS.

Currently these objects are not provided by the source.

To avoid the error "cannot find object" the selection is inactived for now.


jjmurre (approved):
Lijkt me voor nu idd de beste oplossing. In januari maar verder afstemmen.

PR #170, opened 2021-01-07
[DS-394] DAG vergunningen

cvriel:
Included data processing for B&B- and Omzettingsvergunning.
Target table is created based upon Amsterdam schema definition.
Data is processed based upon Change Data Capture (traget table is not dropped)


bheupers (approved):
Ziet er goed uit

jjmurre (approved):
LGTM

PR #177, opened 2021-01-17
[DS-563] add dag_id => dataschema name translation

cvriel:
Some DAG's do not match dag_id with dataschema name.
The reason can be that for one dataset multiple DAG's are implemented i.e. for each table one DAG.
Since the dataschema is the basis for the database permissions, for the situation where dag_id does not match dataschema name, the relation between the dag_id and schema name is held in this constant.

### DIFF
--- a/src/plugins/postgres_permissions_operator.py
+++ b/src/plugins/postgres_permissions_operator.py
@@ -137,6 +137,7 @@ class PostgresPermissionsOperator(BaseOperator):
                     for key in DAG_DATASET.keys():
                         if key in dataset_name:
                             dataset_name = DAG_DATASET[key]
+                            break
 
                     logger.info("set grants for %s", dataset_name)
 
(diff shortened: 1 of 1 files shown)


src/plugins/postgres_permissions_operator.py
```diff
@@ -117,6 +133,11 @@ def execute(self, context=None):  # noqa: C901
 
                 for dataset_name in executed_dags_after_delta:
 
+                    # get real datasetname from DAG_DATASET constant, if dag_id != dataschema name
+                    for key in DAG_DATASET.keys():
+                        if key in dataset_name:
+                            dataset_name = DAG_DATASET[key]
+
```
gkoller, line 140: You might want to add a `break` statement here to terminate the `for` loop once you've found the key. Without a `break` statement this `for` loop could be interpreted as _find the last matching key_, whereas there should only ever be one match.

```python
dataset_name = DAG_DATASET[key]
break
```

cvriel: Goeie, zal ik aanpassen

PR #179, opened 2021-01-19
[DS-CMSA] source sensor data has changed

cvriel:
The source to retreive the sensor data is no longer avialable. The data can be retreived from another endpoint. This also implies different source output. And therefore changed data processing logic. The existing logic is changed to a minimum.


jjmurre (approved):
Lijkt me prima zo.

PR #180, opened 2021-01-19
[DS-569] added jaar from source to data processing

cvriel:
The attribute "jaar" was added in the source to differentiate between same week numbers since the beginning of the year 2021


jjmurre (approved):
LGTM

PR #183, opened 2021-01-21
Fix offset issue in token expiry and made configurable

jjmurre:
Seems there was an offset issue with token expiry. Since new Oauth proxy there are intermittent issues.
This PR should fix that. Furthermore, expiry safety margin has been made configurable.


ngaranko (approved):
LGTM

PR #195, opened 2021-02-17
Use SHARED_DIR as tmp

ngaranko:
Make /tmp directory configurable via env variables.

### DIFF
--- a/src/dags/dag_update.py
+++ b/src/dags/dag_update.py
@@ -5,7 +5,7 @@ from airflow.operators.bash_operator import BashOperator
 from environs import Env
 
 from airflow import DAG
-from common import default_args
+from common import default_args, SHARED_DIR
 
 env = Env()
 
@@ -26,18 +26,18 @@ with DAG(
 
     fetch_dags_from_github = BashOperator(
         task_id="fetch_dags_from_github",
-        bash_command=f"wget {github_url} -O /tmp/repo.zip",
+        bash_command=f"wget {github_url} -O {SHARED_DIR}/repo.zip",
     )
 
     extract_zip = BashOperator(
-        task_id="extract_zip", bash_command="unzip -o /tmp/repo.zip -d /tmp"
+        task_id="extract_zip", bash_command="unzip -o {SHARED_DIR}/repo.zip -d {SHARED_DIR}"
     )
 
     mkdir = BashOperator(task_id="mkdir", bash_command=f"mkdir -p {DAG_SYNC_PATH}")
 
     move_dags = BashOperator(
         task_id="move_dags",
-        bash_command=f"rsync -av /tmp/dataservices-airflow-master/src/dags/ {DAG_SYNC_PATH}",
+        bash_command=f"rsync -av {SHARED_DIR}/dataservices-airflow-master/src/dags/ {DAG_SYNC_PATH}",
     )
 
     # show = PythonOperator(task_id="show", python_callable=doit)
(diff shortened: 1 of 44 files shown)


src/dags/dag_update.py
```diff
@@ -26,18 +26,18 @@ def doit():
 
     fetch_dags_from_github = BashOperator(
         task_id="fetch_dags_from_github",
-        bash_command=f"wget {github_url} -O /tmp/repo.zip",
+        bash_command=f"wget {github_url} -O {SHARED_DIR}/repo.zip",
```
gkoller, line 29: I think this should be `${SHARED_DIR}`. Eg with a `$`

ngaranko: it's an f string, not a bash command yet ;)

gkoller: Oh, no, not here, but you have also `BashOperators` where the string is not a format string. Than it boils down to shell variable expansion and then it does need the `$`

src/dags/dag_update.py
```diff
@@ -26,18 +26,18 @@ def doit():
...
     )
 
     extract_zip = BashOperator(
-        task_id="extract_zip", bash_command="unzip -o /tmp/repo.zip -d /tmp"
+        task_id="extract_zip", bash_command="unzip -o {SHARED_DIR}/repo.zip -d {SHARED_DIR}"
```
gkoller, line 33: Here, is an example. No format string, hence shell variable expansion (I presume), hence the need for a `$`.

jjmurre (approved):
LGTM.  Er lopen nog een paar DAGs via de vsd route. Die doet z'n eigen configuratie van TMPDIR. Die zou je ook nog op SHARED_DIR kunnen baseren. Of wellicht houd je daar al rekening mee? Daar is ook al een SHARED_DIR variabele.

src/vsd/shared/utils/get_objectstore_file.py
```diff
@@ -36,5 +36,5 @@ def get_objectstore_file(location, dir1, connection):
     parser.add_argument("filename")
     parser.add_argument("-s", "--swift_connection")
     args = parser.parse_args()
-    tmpdir = os.getenv("TMPDIR", "/tmp/")
+    tmpdir = os.getenv("SHARED_DIR", "/tmp/")
```
gkoller, line 39: Not in the scope of this PR, but I was wondering how we deal with environment variables. Here we use the `os` module. Most other places use `environs.Env`. We should probably use one way.

With type annotations in mind I'd like to suggest not using `os` or even `environs` but using Pydantic's settings management instead: https://pydantic-docs.helpmanual.io/usage/settings/

src/dags/dag_update.py
ngaranko: Good one, thanks.

src/vsd/shared/utils/get_objectstore_file.py
ngaranko: This one is worth discussion with everyone.

ngaranko:
> LGTM. Er lopen nog een paar DAGs via de vsd route. Die doet z'n eigen configuratie van TMPDIR. Die zou je ook nog op SHARED_DIR kunnen baseren. Of wellicht houd je daar al rekening mee? Daar is ook al een SHARED_DIR variabele.

@jjmurre i checked, most of VSD DAGs will be exedcuted as single step and do not really need shared folder. But maybe it's good to change those as well.

src/vsd/shared/utils/get_objectstore_file.py
gkoller: Agreed. Already posted something on Slack

gkoller (approved):

PR #196, opened 2021-02-18
[DS-588] id field is based on business identification

cvriel:
On request of the user the value for the ID field is based on a business
resource identication. Before that a ID value was generated based on order of load.

The user needs a more steady and reliable key for comparision of different api calls
to detect changes.

### DIFF
--- a/src/dags/importscripts/import_sport.py
+++ b/src/dags/importscripts/import_sport.py
@@ -1,0 +1,99 @@
+import csv as csv
+from hashlib import blake2s
+import json
+import logging
+from typing import List, Tuple, TypedDict
+
+# define logger for output to console
+logger = logging.getLogger(__name__)
+
+
+def get_dialect(file: str):
+    """determine dialect of file"""
+    with open(file) as data:
+        return csv.Sniffer().sniff(data.readline())
+
+
+def get_header(file: str) -> str:
+    """Return header row of file"""
+    dialect = get_dialect(file)
+    with open(file, "r") as src:
+        data = csv.reader(src, dialect=dialect)
+        return next(data)
+
+
+def make_hash(composite_values: List, digest_size: int = None) -> int:
+    """The blake2s algorithm is used to generate a single hased value for source
+    composite values that uniquely identify a row.
+    In case the source itself doesn't provide a single solid identification as key.
+
+    Note:
+    The default digest size is for now set to 3 bytes, which is equivant
+    to ~ 10 karakters long.
+    Because the id column is preferably of type int, the hased value is converted
+    from hex to int.
+    """
+    digest_size = 3 if not digest_size else digest_size
+    return int(
+        blake2s("|".join(composite_values).encode(), digest_size=digest_size).hexdigest(), base=16
+    )
+
+
+def read_data(file: str, composite_key: Tuple):
+    """Read the data from csv file and contain it into a data class"""
+    dialect = get_dialect(file)
+    header = get_header(file)
+    # setup the generic TypeDict data mold
+    DataObject = TypedDict("DataObject", {col: str for col in header})
+    with open(file, "r") as src:
+        data = csv.reader(src, dialect=dialect)
+        for line in data:
+            # construct lines into dictionary format, to proces into the generic data mold
+            row_dict = {i[0]: i[1] for i in zip(header, line)}
+            row_object = DataObject(object_=DataObject(row_dict), composite_key=composite_key)
+            yield row_object
+
+
+def generate_unique_id(file: str, composite_key: Tuple, digest_size: int = None) -> dict:
+    """Generate unique hased id from composite values"""
+    for row in read_data(file, composite_key):
+        composite_values = []
+        for column in row["composite_key"]:
+            composite_values.append(row["object_"].get(column, None))
+        try:
+            row["object_"]["id"] = make_hash(composite_values, digest_size=digest_size)
+            yield row["object_"]
+        except TypeError as err:
+            logger.error("%s cannot be retrieved from %s: %s", column, file, err)
+            continue
+
+
+def add_unique_id_to_csv(file: str, composite_key: Tuple, digest_size: int = None):
+    """save enriched data to csv for further processing in DAG"""
+    data = generate_unique_id(file, composite_key=composite_key)
+    header = next(data)  # get the modified header including the new id column
+    rows = list()
+    for row in data:
+        rows.append(row.values())
+    with open(file, "w") as f:
+        write = csv.writer(f, dialect=csv.unix_dialect)
+        write.writerow(header)
+        write.writerows(rows)
+
+
+def add_unique_id_to_geojson(file: str, composite_key: Tuple, digest_size: int = None):
+    """save enriched data to geojson for further processing in DAG"""
+    with open(file, "r+") as f:
+        data = json.load(f)
+        for rows in data["features"]:
+            key_values = []
+            for key in composite_key:
+                # openbareruimte geojson has a different structure
+                if "openbaresportplek" in file:
+                    key_values.append(rows[key])
+                else:
+                    key_values.append(rows["properties"][key])
+                rows["properties"]["id"] = make_hash(["".join(str(key_values))])
+        f.seek(0)  # set reader back to line 0
+        f.truncate()
+        f.write(json.dumps(data))
--- a/src/dags/importscripts/import_sport.py
+++ b/src/dags/importscripts/import_sport.py
@@ -2,28 +2,28 @@ import csv
 from hashlib import blake2s
 import json
 import logging
-from typing import List, Tuple, TypedDict, Generator, Type
+from typing import List, Tuple, Iterator, Type, Dict, Any, Optional
 
 
 # define logger for output to console
 logger = logging.getLogger(__name__)
 
 
-def get_dialect(file: str) -> Type:
+def get_dialect(file: str) -> Type[csv.Dialect]:
     """determine dialect of file
 
     Args:
         file: the path/to/file.ext that will be used to determine it's dialect
 
     Returns:
-        A subclass dialect of Sniffer
+        A subclass Dialect of Sniffer
 
     """
     with open(file) as data:
         return csv.Sniffer().sniff(data.readline())
 
 
-def get_header(file: str) -> List:
+def get_header(file: str) -> List[str]:
     """Return header row of file
 
     Args:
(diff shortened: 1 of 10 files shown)


gkoller:
Like what you're doing here (`import_sport.py`). Lot's of type annotations, lots of docstrings. Nice!

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+import json
+import logging
+from typing import List, Tuple, TypedDict, Generator, Type
+
+
+# define logger for output to console
+logger = logging.getLogger(__name__)
+
+
+def get_dialect(file: str) -> Type:
```
gkoller, line 12: Some type annotations can be a bit more specific. You're returning a `Dialect` type here, hence `-> Type[Dialect]`

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Returns:
+        A subclass dialect of Sniffer
+
+    """
+    with open(file) as data:
+        return csv.Sniffer().sniff(data.readline())
+
+
+def get_header(file: str) -> List:
```
gkoller, line 26: `-> List[str]`

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+        a list of columns in string
+
+    """
+    dialect = get_dialect(file)
+    with open(file, "r") as src:
+        data = csv.reader(src, dialect=dialect)
+        return next(data)
+
+
+def make_hash(composite_values: List, digest_size: int = None) -> int:
```
gkoller, line 42: `composite_values: List[str]`

And either: 
* `digest_size: int = 3` and remove the inline if statement in de function body (preferred) 
* or `digest_size: Optional[int] = None` and keep the inline if statement. (not recommend as you would need an explicit test for `is not None` to satisfy MyPy before passing `digest_size` to `blake2s`

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+    """The blake2s algorithm is used to generate a single hased value for source
+    composite values that uniquely identify a row.
+    In case the source itself doesn't provide a single solid identification as key.
+
+    Args:
+        composite_values: a concatenated string of record values based on
```
gkoller, line 48: The concatenation happens in the function body, hence the parameter is not yet concatenated.

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+                        which columns unqiuely identifiy a record
```
gkoller, line 49: uniquely

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+        digest_size: size to set the max bytes to use for the hash
+
+    Returns:
+        a hased value of the composite values in int
+
+    Note:
+        The default digest size is for now set to 3 bytes, which is equivant
```
gkoller, line 56: equivalent

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+        to ~ 10 karakters long.
```
gkoller, line 57: Actually a 3 byte `digest_size`  creates a 6 character long hex digest. Reason being that each hex digit encodes 4 bits (16 values). Each byte has 2 x 4 bits. Hence 3 bytes creates a 2 x 6 hex digit digest.

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+        Because the id column is preferably of type int, the hased value is converted
+        from hex to int.
+    """
+    digest_size = 3 if not digest_size else digest_size
+    return int(
+        blake2s("|".join(composite_values).encode(), digest_size=digest_size).hexdigest(), base=16
```
gkoller, line 63: You are converting a digest in bytes to hex and then from hex to int. This can be done in one step:

```python
int.from_bytes(blake2s(...).digest())
```

gkoller: `digest()` returns the internal byte representation (no conversion). `hexdigest()` does do a conversion.

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+
+
+def read_data(file: str, composite_key: Tuple) -> Generator:
```
gkoller, line 67: `composite_key: Tuple[str]`

A `Generator` is [specified](https://docs.python.org/3.8/library/typing.html?highlight=generator#typing.Generator) as follows: `Generator[YieldType, SendType, ReturnType]`. If you are only yielding values you are better of using `Iterator`. 

First move the `DataObject` TypedDict definition out of the function body and specify `Iterator[DataObject]` as the return type of the function. 

Next use [DictReader](https://docs.python.org/3.8/library/csv.html?highlight=dictreader#csv.DictReader) and simplify the for loop as follows:

```python
for line in data:
    yield cast(DataObject, line)
```
This means you can also do away with `header = get_header(file)`

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+    """Read the data from csv file and contain it into a data class
+
+    Args:
+        file: the path/to/file.ext that will be read
+        composite_key: Tuple of column name(s) that unqiuely identifiy a record
```
gkoller, line 72: uniquely ;-)

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+        from hex to int.
+    """
+    digest_size = 3 if not digest_size else digest_size
+    return int(
+        blake2s("|".join(composite_values).encode(), digest_size=digest_size).hexdigest(), base=16
+    )
+
+
+def read_data(file: str, composite_key: Tuple) -> Generator:
+    """Read the data from csv file and contain it into a data class
```
gkoller, line 68: I'm sorry for nitpicking, but either you should use a real data class (as suggested by your docstring), or if you want to stick to the `TypedDict`, leave _and contain it into a data class_ out of the docstring. What the function returns is already specified in the return type hence you can be brief in your docstring.

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Args:
+        file: the path/to/file.ext that will be read
+        composite_key: Tuple of column name(s) that unqiuely identifiy a record
+
+    Returns:
+        a generator containing the rows of the file plus it's header and the defined composite key
```
gkoller, line 75: It does not return a Generator. It yields row values. Replace `Returns` with `Yields` Eg.

```python
Yields:
   The next row from the data file.
```
See also: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html (though no need to specify the type as they do there as we are using type annotations for that)

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+    header = get_header(file)
+    # setup the generic TypeDict data mold
+    DataObject = TypedDict("DataObject", {col: str for col in header})
+    with open(file, "r") as src:
+        data = csv.reader(src, dialect=dialect)
+        for line in data:
+            # construct lines into dictionary format, to proces into the generic data mold
+            row_dict = {i[0]: i[1] for i in zip(header, line)}
+            row_object = dict(object_=DataObject(row_dict), composite_key=composite_key)
+            yield row_object
```
gkoller, line 88: The `csv` mode has the [DictReader](https://docs.python.org/3.8/library/csv.html?highlight=dictreader#csv.DictReader). It does what you are trying to do here. By using `DictReader` you can simplify the for loop as follows:

```python
for line in data:
    yield cast(DataObject, line)
```
The `cast` is necessary as `DictReader` returns untyped dicts.

This means you can also do away with `header = get_header(file)`

src/dags/importscripts/import_sport.py
gkoller: Oeps, you can't move the `DataObject` definition out of the function body as its definition is dynamically created. I wouldn't do that as dynamically created data types don't do much for static analysis (or for you IDE of for other programmers.)

Just use `Iterator[Dict[str, Any]]`

src/dags/importscripts/import_sport.py
gkoller: See also https://github.com/Amsterdam/dataservices-airflow/pull/196/files#r580020020

`DataObject` might nog be necessary. If you use `Iterator[Dict[str, Any]` as your function return type. You might need still need a cast here, but now it becomes:
```python
yield cast(Dict[str, Any], line)
```

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def generate_unique_id(file: str, composite_key: Tuple, digest_size: int = None) -> Generator:
```
gkoller, line 91: * `composite_key: Tuple[str]`
* `digest_size: Optional[int] = None`
* `-> Iterator[Dict[str, Any]]`

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+    Returns:
+        a generator containing the rows plus the unique ID value for the added ID column
+
+    """
+    for row in read_data(file, composite_key):
+        composite_values = []
+        for column in row["composite_key"]:
+            composite_values.append(row["object_"].get(column, None))
+        try:
+            row["object_"]["id"] = make_hash(composite_values, digest_size=digest_size)
```
gkoller, line 108: Because `make_hash`'s `digest_size` is no longer `Optional` (if you've followed my suggestions) this now becomes:

```python
options = {"digest_size": digest_size} if digest_size else {}
row["object_"]["id"] = make_hash(composite_values, **options)
```

src/dags/importscripts/import_sport.py
```diff
@@ -0,0 +1,163 @@
...
+    rows = list()
+    for row in data:
+        rows.append(row.values())
+    with open(file, "w") as f:
+        write = csv.writer(f, dialect=csv.unix_dialect)
+        write.writerow(header)
+        write.writerows(rows)
+
+
+def add_unique_id_to_geojson(file: str, composite_key: Tuple, digest_size: int = None) -> None:
```
gkoller, line 138: `composite_key: Tuple[str]`

You can remove `digest_size` as you are not using it.

src/dags/importscripts/import_sport.py
gkoller: On second thought I'm not sure my remark is valid. After all you are concatenating things on line 160 in you call to `make_hash`

src/dags/importscripts/import_sport.py
cvriel: Aangepast

src/dags/importscripts/import_sport.py
cvriel: Aangepast

src/dags/importscripts/import_sport.py
cvriel: Aangepast, gekozen voor digest_size: int = 3

src/dags/importscripts/import_sport.py
cvriel: Dit is een dubbele concatenate, heb deze over het hoofd gezien. Ik laat die in de hash functie staan.

src/dags/importscripts/import_sport.py
cvriel: Oeps, typo

src/dags/importscripts/import_sport.py
cvriel: Oeps, typo

src/dags/importscripts/import_sport.py
cvriel: Je hebt gelijk, ik had eerst een digist_size van 5. Wat is gelijk aan 10 karakters. De tekst is niet aangepast. Van een 4 bits codering naar bytes is i.d.d. 3 / 0,5 = 6 karakters

src/dags/importscripts/import_sport.py
cvriel: Aangepast

src/dags/importscripts/import_sport.py
cvriel: At least the typo is consistent ;-)

src/dags/importscripts/import_sport.py
cvriel: Eens, aangepast

src/dags/importscripts/import_sport.py
cvriel: Aangepast

src/dags/importscripts/import_sport.py
cvriel: Aangepast

src/dags/importscripts/import_sport.py
cvriel: Oeps, foutje

PR #203, opened 2021-03-02
Pipeline procesverbaal verkiezingen

cvriel:
This dag is just run for a couple of days. To collect all mandatory publications
after a election. It is supposed to run every 10 minutes. When all documents are
uploaded en processed in the database (as a URI) then the dag can be set to inactive.

### DIFF
--- a/src/dags/importscripts/import_processenverbaalverkiezingen.py
+++ b/src/dags/importscripts/import_processenverbaalverkiezingen.py
@@ -1,4 +1,5 @@
 import csv
+from dataclasses import dataclass
 import ftplib
 from more_ds.network.url import URL
 from os import path as ospath
@@ -16,7 +17,15 @@ class Data(TypedDict):
     id: str
 
 
-class FTPListing:
+@dataclass
+class ListElement:
+    """Datastructure for listing elements providing more context"""
+
+    relative_path: str
+    files: str
+
+
+class ObjectStoreListing:
     """This class is used for traversing a given folder base on a FTP
     server (objectstore) using BFS (Breadth-First search) algorithm.
     """
@@ -104,19 +113,31 @@ def save_data(
     data_to_save: List = []
     connection = ftplib.FTP(host=host)
     connection.login(user=user, passwd=passwd)
-    get_listing = FTPListing(connection)
+    get_listing = ObjectStoreListing(connection)
 
     for data in get_listing.traverse_folder(startfolder):
-        for file in data[2]:
+        resultlist = ListElement(relative_path=data[0], files=data[2])
 
-            verkiezingsjaar = data[0].split("/")[1]
+        for file in resultlist.files:
+
             volgnummer = file.split(".")[0]
-            uri = URL(prefix_url) / data[0] / file
             documentnaam = file.split(".")[1]
             stemlocatie = file.split(".")[2]
+            uri = URL(prefix_url) / resultlist.relative_path / file
+            verkiezingsjaar = resultlist.relative_path.split("/")[1]
+
+            try:
+                verkiezingsjaar_int = int(verkiezingsjaar)
+            except ValueError as err:
+                raise ValueError(
+                    f"""Verkiezingsjaar is not a number. Check the folder
+                name where the files 'processenverbaal' are located: {resultlist.relative_path}
+                The folder name must be set as YYYY as year of election.
+                """
+                ) from err
 
             metadata = Data(
-                verkiezingsjaar=verkiezingsjaar,
+                verkiezingsjaar=verkiezingsjaar_int,
                 volgnummer=volgnummer,
                 uri=uri,
                 documentnaam=documentnaam,
--- a/src/dags/importscripts/import_processenverbaalverkiezingen.py
+++ b/src/dags/importscripts/import_processenverbaalverkiezingen.py
@@ -52,12 +52,12 @@ class ObjectStoreListing:
 
         """
 
-        list_, dirs, files = [], [], []
         try:
             self.connection.cwd(_path)
         except Exception:
             return [], []
         else:
+            list_, dirs, files = [], [], []
             self.connection.retrlines("LIST", lambda x: list_.append(x.split()))
             for info in list_:
                 type, name = info[0], info[-1]
@@ -146,10 +146,9 @@ def save_data(
             )
             data_to_save.append(metadata)
 
-    header = Data.__annotations__.keys()
-    data = [row.values() for row in data_to_save]
+    fieldnames = Data.__annotations__.keys()
 
     with open(output_file, "w") as f:
-        write = csv.writer(f, dialect=csv.unix_dialect)
-        write.writerow(header)
-        write.writerows(data)
+        writer = csv.DictWriter(f, fieldnames=fieldnames, dialect=csv.unix_dialect)
+        writer.writeheader()
+        writer.writerows(data_to_save)
--- a/src/dags/importscripts/import_processenverbaalverkiezingen.py
+++ b/src/dags/importscripts/import_processenverbaalverkiezingen.py
@@ -1,0 +1,134 @@
+import csv
+import ftplib
+from more_ds.network.url import URL
+from os import path as ospath
+from typing import List, TypedDict, Tuple, Optional, Generator
+
+
+class Data(TypedDict):
+    """Datastructure processenverbaal verkiezingen"""
+
+    verkiezingsjaar: int
+    volgnummer: str
+    uri: str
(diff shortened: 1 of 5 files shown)


src/dags/importscripts/import_processenverbaalverkiezingen.py
```diff
@@ -0,0 +1,134 @@
...
+    def __init__(self, connection: ftplib.FTP) -> None:
+        """Initialize
+
+        Args:
+            Connection: Hold the connection to the objectstore
+
+        """
+        self.connection = connection
+
+    def list_dirs_and_files(self, _path: str) -> Tuple[List[str], Optional[List[str]]]:
```
gkoller, line 33:
```python
-> Tuple[List[str], List[str]]
```
The second element of the returned tuple (`files`) will never be `None`. In the body of the function you initialize it to `[]` and you are only appending to it. Hence `files` will never be `None`

src/dags/importscripts/import_processenverbaalverkiezingen.py
```diff
@@ -0,0 +1,134 @@
...
+            self.connection.retrlines("LIST", lambda x: list_.append(x.split()))
+            for info in list_:
+                type, name = info[0], info[-1]
+                if type.startswith("d"):
+                    dirs.append(name)
+                else:
+                    files.append(name)
+            return dirs, files
+
+    def traverse_folder(self, path: str = "/") -> Generator:
```
gkoller, line 61:
```python
-> Iterator[Tuple[str, List[str], List[str]]]
```
Though you might want to consider using a `dataclass` for this.  After all in the function `save_data` you are referring to `data[0]`, `data[1]` and `data[2]`. Without looking at the function body of `traverse_folder` it is not immediately clear what each element refers to. Hence you might want to consider this:

```python
from dataclasses import dataclass, field
from typing import List

@dataclass
class DirList:
    path: str
    dirs: List[str] = field(default_factory=list)
    files: List[str] = field(default_factory=list)
```
And use that in your `list_dirs_and_files` and traverse_folder` methods.

Anyway with `DirList` in place code in `save_data` would change from something like:

```python
    for data in get_listing.traverse_folder(startfolder):
        for file in data[2]:
```
to
```python
    for data in get_listing.traverse_folder(startfolder):
        for file in data.files:
```
See the difference between `data[2]` versus `data.files`?

In addition `data[0]` becomes `data.path` (and `data[1]` becomes `data.dirs`)

src/dags/importscripts/import_processenverbaalverkiezingen.py
```diff
@@ -0,0 +1,134 @@
...
+        output_file: name of .csv file to save
+
+    Executes:
+        Stores file URL's and it's metadata to .csv file
+
+    Notes:
+        The filenames are meaningful. It contains it's metadata. For example:
+        `001.procesverbaaltk21.Amstel1.pdf` conceals <volgnummer>.<documentnaam>.<stemlocatie>.pdf
+        Furthermore, the files are located in a folder which name
+        represents it's election year i.e. 2021
```
gkoller, line 101: I like the extensive documentation! :+1:

src/dags/importscripts/import_processenverbaalverkiezingen.py
```diff
@@ -0,0 +1,134 @@
...
+
+    verkiezingsjaar: int
+    volgnummer: str
+    uri: str
+    documentnaam: str
+    stemlocatie: str
+    id: str
+
+
+class FTPListing:
```
jjmurre, line 19: Is this using a special FTP server? Or just the objectstore using the FTP protocol?

src/dags/importscripts/import_processenverbaalverkiezingen.py
gkoller: Having said that you might want to consider using a dataclass. See https://github.com/Amsterdam/dataservices-airflow/pull/203/files#r585502118

src/dags/importscripts/import_processenverbaalverkiezingen.py
```diff
@@ -0,0 +1,134 @@
...
+                volgnummer=volgnummer,
+                uri=uri,
+                documentnaam=documentnaam,
+                stemlocatie=stemlocatie,
+                id=verkiezingsjaar + volgnummer,
+            )
+            data_to_save.append(metadata)
+
+    header = Data.__annotations__.keys()
+    data = [row.values() for row in data_to_save]
```
gkoller, line 129: You're doing what [csv.DictWriter](https://docs.python.org/3.8/library/csv.html?highlight=dictwriter#csv.DictWriter) does for you automatically.

In other words. Use `csv.DictWriter` and you can write each dict directly (in a loop).

gkoller: Should you choose to stick to your current implement than at least use `newline=""` in your call to `open`.

See also: https://docs.python.org/3.8/library/csv.html?highlight=dictwriter#id3 (the footnote of that document)

gkoller (approved):

src/dags/importscripts/import_processenverbaalverkiezingen.py
```diff
@@ -0,0 +1,134 @@
...
+
+        Args:
+            _path: The path for looking for files and directories (within the same path)
+
+        Returns:
+            list of found dictories and files
+
+        """
+
+        list_, dirs, files = [], [], []
```
lbam, line 46: I'd put these closer to their use site, inside the else block.

src/dags/importscripts/import_processenverbaalverkiezingen.py
```diff
@@ -0,0 +1,134 @@
...
+            list of all files incl its path until all directories are depleted
+
+        """
+        dirs, files = self.list_dirs_and_files(path)
+        yield path, dirs, files
+        for name in dirs:
+            path = ospath.join(path, name)
+            yield from self.traverse_folder(path)
+            self.connection.cwd("..")
+            path = ospath.dirname(path)
```
lbam, line 78: This is a pre-order traversal, not breadth-first.

src/dags/importscripts/import_processenverbaalverkiezingen.py
lbam: it's -> its

src/dags/importscripts/import_processenverbaalverkiezingen.py
cvriel: :+1:

src/dags/importscripts/import_processenverbaalverkiezingen.py
cvriel: :+1: interesting thanks! First time using the dataclass, it's very handy. I wonder if this could also work with a TypedDict. Making it duck typing proof if there is a use case for it off course, next time maybe!

src/dags/importscripts/import_processenverbaalverkiezingen.py
cvriel: :+

src/dags/importscripts/import_processenverbaalverkiezingen.py
cvriel: You are right :+1:  This is indeed the Objectstore. To me more precise I will rename it to ObjectStoreListing

src/dags/importscripts/import_processenverbaalverkiezingen.py
cvriel: I like the DictWriter approach more, I adopted your suggestion

src/dags/importscripts/import_processenverbaalverkiezingen.py
cvriel: Adjusted

src/dags/importscripts/import_processenverbaalverkiezingen.py
cvriel: Actually it does, it gets the first level dirs, then goes to the next level etc, and ends at the leaf folders (lowest level). Like so:
In this case 2022 had no child folders. But is processed at the same time als 2021. As it's the top level.

1) ['2021', '2022']
2) ['2021_1', '2021_2']
3) ['2021_1_1']

lbam: 2022 having no children is cheating ;)

If the hierarchy has files

```
a1/a2/a3/a4
b1/b2/b3/b4
```

(and intermediate directories), then your algorithm either produces a4 before b3, or b4 before a3. BFS never does that.

PR #208, opened 2021-03-04
Accomodate the use of shortnames in schemas

jjmurre:
In the previous implementation, the dataset schema was only needed in the
HttpGobOperator. The names of tables involved in the import process
were having fixed names and were derived from the names of the
graphql definition files.

In the new situation, table (and column) names can be overridden from the
amsterdam schema. So, the dataset schema information is needed
throughout the whole DAG (e.g. for table renaming).

It is bad practice to load the schema information directly in the
dag module, because this module is reloaded numerous times by
Airflow. All processing should be done exclusively in operators.

So, a new operator had been added at the start of the DAG
that reads the relevant schema and put this information on XCom.

Operators later in the DAG that need information about table names
can fetch this information from XCom.

This involved the modification of several operators more
downstream in the DAG. Those operator were relying on table
names entering the operator during construction.
However, when a `shortname` is used in the schema definition,
the tablename needs to be obtained from XCom.

So the relevant operators have been modified to accommodate this
behaviour.

To keep the code as DRY as possible, this XCom-related behaviour
has been put in a mixin class (XComAttrAssignerMixin).

To keep the code generic, a function can be injected into these
operators that is able to modify the relevant operator attributes
(usually attributes holding table names).


src/dags/gob.py
```diff
@@ -42,63 +59,94 @@ def create_gob_dag(is_first, gob_dataset_name, gob_table_name):
...
+
+        # 2. Create Dataset info to put on the xcom channel for later use
+        # by operators
+        create_dataset_info = PythonOperator(
+            task_id=f"mkinfo_{dataset_table_id}",
+            python_callable=_create_dataset_info,
+            op_args=(gob_dataset_id, gob_table_id),
+        )
+
+        def init_assigner(o: Any, x: Any) -> None:
```
gkoller, line 106: Maybe this can be a little more specific. Setting the attribute always seems to happen on a `PostgresOperator` derived class. Hence you'd probably want at least:
```python
def init_assigner(o: PostgresOperator, x: Any) -> None:
   ...
```
However MyPy will not like that, as dynamically setting an attribute on a class that doesn't define that attribute is technically not type correct. That is, unless you define `__setattr__` specifically. But that means defining an intermediate base class:

```python
class AttrPostgresOperator(PostgresOperator):
    """"Allow any property to be get and set in MyPy compatible way."""
    def __setattr__(self, name: str, value: Any) -> None:
        super().__setattr__(name, value)

    def __getattr__(self, name: str) -> Any:
        return super().__getattribute__(name)

class PostgresTableInitOperator(AttrPostgresOperator):
    ...
```
Which in turn means that the type annotation on `init_assigner` becomes:
```python
def init_assigner(o: AttrPostgresOperator, x: Any) -> None:
   ...
```
That might be a bit too much. Anyway got this from https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html#when-you-re-puzzled-or-when-things-are-complicated (see `class A` and the comments)

gkoller (approved):

PR #222, opened 2021-03-11
Fix for log remover DAG

cvriel:
The LOG remover DAG doesn't work since the Airflow upgrade.

One of the changes due to the new version was the
run_dag.conf.maxLogAgeInDays is not available anymore. But it is not needed any way. Because the window for holding logs is already definied in an environment variable in the docker-compose.yaml file. It is currently set on 90 days.


ngaranko:
Nice! Thanks, was planning to fix that today, but got dragged into other things.

ngaranko (approved):

PR #228, opened 2021-03-24
WIOR changes due production data source

cvriel:
In the source production environment the geometry can be various. From polygon to lines and points.
Therefor the logic is changed accordingly


jjmurre (approved):
👍🏼

PR #229, opened 2021-03-24
AB#6139

cvriel:
Processenverbaalverkiezingen was not yet added to the DAG_DATASET dictionary for sets where the DAG_ID is not the same as the schema name. Therefor the grants could not be set properly.


jjmurre (approved):
👍🏼

PR #233, opened 2021-03-30
AB#6372 changed http_fetch_operator for maps.amsterdam.nl and sport update

cvriel:
Currently one of the data sources of sport is failing. To cope with the situation
the http_fetch_operator is changed so it can handle the maps.amsterdam.nl source.

Also in the sports datasets an extra step for cleaning up the data is added (on request of the user).

### DIFF
--- a/src/dags/sport.py
+++ b/src/dags/sport.py
@@ -14,6 +14,7 @@ from provenance_rename_operator import ProvenanceRenameOperator
 from typeahead_location_operator import TypeAHeadLocationOperator
 from pgcomparator_cdc_operator import PgComparatorCDCOperator
 from sqlalchemy_create_object_operator import SqlAlchemyCreateObjectOperator
+from postgres_permissions_operator import PostgresPermissionsOperator
 
 from common import (
     default_args,
@@ -21,6 +22,7 @@ from common import (
     DATAPUNT_ENVIRONMENT,
     MessageOperator,
     SHARED_DIR,
+    quote_string,
 )
 
 from postgres_check_operator import (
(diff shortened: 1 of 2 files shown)


src/dags/sport.py
```diff
@@ -14,13 +14,15 @@
 from typeahead_location_operator import TypeAHeadLocationOperator
```
ngaranko, line 14: This file does not really belong to `http_fetch_operator` ;)

src/plugins/http_fetch_operator.py
```diff
@@ -71,7 +71,7 @@ def __init__(
 
         super().__init__(*args, **kwargs)
 
-    def execute(self, context: Dict) -> None:
+    def execute(self, context: Dict) -> None:  # noqa
```
ngaranko, line 74: Try avoiding plain `# noqa` comments. 
Either fix the line or specify error code to be ignored explicitly: `# noqa: E231`. (note semicolon between `noqa` and code).

src/dags/sport.py
cvriel: That's true. Because the sport dataset benefits from the http_fetch_operator (one source of the sports dataset is not loaded without the workarround) and there is also a request for some minor adjustements, I combined the two in one PR. Maybe I should have split them up for a cleaner track

src/plugins/http_fetch_operator.py
cvriel: Thanks, I will make it more specific. I used noqa - in this case - to strictly focus on the workarround without getting distracted by earlier code that is flagged by flake8

ngaranko (approved):

PR #240, opened 2021-04-13
Prevent head-tail collision of estafette-run DAG series

jjmurre:
The trigger operator that triggers the next dag that is running serialized (estafette run) tries to start a DAG that is still running (head-tail collision). This change adds an extra check to prevent it.

Furthermore, additional documentation and missing type info has been added.

### DIFF
--- a/src/plugins/dynamic_dagrun_operator.py
+++ b/src/plugins/dynamic_dagrun_operator.py
@@ -1,30 +1,57 @@
+from __future__ import annotations
+from datetime import datetime
 from sqlalchemy import not_
+from typing import Optional, Any
 from airflow.settings import Session
-from airflow.models.dag import DagModel
-from airflow.utils.decorators import apply_defaults
+from airflow.models.dag import DagModel, DagRun
 from airflow.operators.dagrun_operator import TriggerDagRunOperator
+from airflow.utils.decorators import apply_defaults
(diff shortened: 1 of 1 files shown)


src/plugins/dynamic_dagrun_operator.py
```diff
@@ -1,30 +1,57 @@
...
             trigger_dag_id=None,
             execution_date=None,
             *args,
             **kwargs,
         )
+        assert len(dag_id_prefix) > 0, "A dag_id_prefix is mandatory to use this Operator."
         self.dag_id_prefix = dag_id_prefix
 
-    def execute(self, context):
+    def execute(self, context: dict[str, Any]) -> None:
```
gkoller, line 54: This is only supported from Python 3.9 onwards: https://www.python.org/dev/peps/pep-0585/

For Python 3.8 you need to use `Dict[str, Any]` eg with a capital `D`

gkoller (approved):
LGTM

src/plugins/dynamic_dagrun_operator.py
jjmurre: But, from the same PEP: Starting with Python 3.7, when `from __future__ import annotations` is used, function and variable annotations can parameterize standard collections directly. Or, is that considered bad practice?

PR #242, opened 2021-04-14
WIOR (werken in openbare ruimte)

cvriel:
On request product owner following changes where made:

- Removal of fields that are not needed (i.e. indication 'kleinwerk')
- Removal of rows that are not needed (status intake)


jjmurre (approved):
LGTM

PR #243, opened 2021-04-14
RDW structure change

cvriel:
The main resource 'rdw voertuig' is flanked by two child tables 'assen' and 'brandstof'. In the schema definition the 'assen' and 'brandstof' data is 'embedded' as an object (array) within the API result.


jjmurre (approved):
Mooi!

PR #247, opened 2021-05-03
[WIOR] added an upload step to store the data in the objectstore

cvriel:
Currently the WIOR (werken in openbare ruimte) data is partially served and processed in the database. To store the complete 'raw' data, its uploaded after download into the objectstore. The files are flagged with the current day so a history of data processing can be registred and serviced if needed to other consumers.

### DIFF
--- a/src/plugins/swift_hook.py
+++ b/src/plugins/swift_hook.py
@@ -1,9 +1,14 @@
 import ntpath
 from contextlib import contextmanager
+from datetime import datetime, timezone, timedelta, tzinfo
+from dateutil import tz
 from pathlib import Path
 from airflow.hooks.base import BaseHook
 from airflow.exceptions import AirflowException
 from swiftclient.service import SwiftService, SwiftError, SwiftUploadObject
+from typing import Optional, Iterator, Dict
+
+to_zone: Optional[tzinfo] = tz.gettz("Europe/Amsterdam")
 
 
 class SwiftHook(BaseHook):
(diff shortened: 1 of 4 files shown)


jjmurre (approved):
Ziet er goed uit. En gelijk een goede toevoeging aan de Swift operator!

src/plugins/swift_hook.py
```diff
@@ -28,7 +39,19 @@ def connection(self):
...
+        """Returns the items in the objectstore folder (container)
+
+        Args:
+            container (str): The objectstore folder to retreive the files
+
+        Raises:
+            AirflowException: Container cannot be listed
+
+        Yields:
+            Iterator[str]: Iterator[str]: Found file in given objectstor folder (container)
```
gkoller, line 52: With type annotations used in the function signature you don't need to repeat them in the docstring. 

Also there is a mismatch between what the function says it returns `Iterator[Dict]` versus what the docstring says it returns `Iterator[str]`

src/plugins/swift_hook.py
```diff
@@ -100,3 +145,75 @@ def upload(self, container, local_path, object_id):  # noqa C901
             except SwiftError as e:
                 self.log.error(e.value)
                 raise AirflowException("Failed to upload file") from e
+
+    def identify_files_not_in_timewindow(
+        self, container: str, time_window_in_days: int
+    ) -> Iterator[list]:
```
gkoller, line 151: I think this should be `Iterator[str]`. That is if `yield file["name"]` yields a `str` and not a `list`

PR #251, opened 2021-05-19
snake case was not applied to swap operator

cvriel:


jjmurre (approved):
Altijd weer die casing :-)   Lijkt me prima zo.

PR #255, opened 2021-05-20
AB#6837 table creation solely based on schema

cvriel:
The existing DAG beheerkaartbasis was lacking the table creation based on the schema defintion. Also the whishes of the users were to create the database objects and it's comments on the schema defintion structure and it's description fields.

The existing DAG was enriched by the extra step of creating the database objects. However, the source data are SQL copy statements that reference a different schema then public. So the tables to be created had to be created in the specified schema as the source copy statements dictated (the PTE schema). The schema option was already available as a paramater in the create objects method from schema-tools. So it could be set.

Also temorary database tables names, as stated in the copy SQL statements from source file, was an option to set. This way the data structure could be created solely from the schema and the source SQL copy statements could be excuted to insert the data.

### DIFF
--- a/src/dags/beheerkaart.py
+++ b/src/dags/beheerkaart.py
@@ -76,7 +76,7 @@ with DAG(
         for data_table_name, db_table_name in tables.items()
     ]
 
-    # 3. load the dump file
+    # 4. load the dump file
     swift_load_task = SwiftLoadSqlOperator(
         task_id="swift_load_task",
         container="Dataservices",
@@ -87,7 +87,7 @@ with DAG(
         db_target_schema="pte",
     )
 
-    # 4. Make the provenance translations
+    # 5. Make the provenance translations
     provenance_renames = ProvenanceRenameOperator(
         task_id="provenance_renames",
         dataset_name=dataset_name,
@@ -97,13 +97,13 @@ with DAG(
         rename_indexes=True,
     )
 
-    # 5. Swap tables to target schema public
+    # 6. Swap tables to target schema public
     swap_schema = SwapSchemaOperator(task_id="swap_schema", dataset_name=dataset_name)
 
-    # 6. Create temporary directory
+    # 7. Create temporary directory
     mkdir = BashOperator(task_id="mkdir", bash_command=f"mkdir -p {tmp_dir}")
 
-    # 7. Create geopackage
+    # 8. Create geopackage
     create_geopackage = BashEnvOperator(
         task_id="create_geopackage",
         env={},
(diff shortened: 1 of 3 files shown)


src/dags/beheerkaart.py
```diff
@@ -104,11 +128,13 @@
...
+    grant_db_permissions = PostgresPermissionsOperator(task_id="grants", dag_name=dataset_name)
 
 # FLOW
-slack_at_start >> drop_tables >> swift_load_task >> provenance_renames >> swap_schema >> mkdir >> create_geopackage >> zip_geopackage >> upload_data >> grant_db_permissions  # noqa
+slack_at_start >> drop_tables >> create_tables
+
+for table in create_tables:
+    table >> swift_load_task
+
+swift_load_task >> provenance_renames >> swap_schema >> mkdir >> create_geopackage >> zip_geopackage >> upload_data >> grant_db_permissions  # noqa
```
jjmurre: Wellicht hier nog even haakjes omheen, dat leest wat prettiger, omdat het vertikaal geformatteerd wordt en is het beter ivm. diffen.

src/plugins/sqlalchemy_create_object_operator.py
```diff
@@ -103,16 +107,27 @@ def execute(self, context: Dict[str, Any]) -> None:
...
+        # setup the database schema for the database connection
+        if self.pg_schema is not None:
+            engine = _get_engine(
+                self.db_conn,
+                pg_schemas=[
+                    self.pg_schema,
+                ],
+            )
+        else:
+            engine = _get_engine(self.db_conn)
```
jjmurre, line 119: Dit zou nog wat korter kunnen met:

    kwargs = {"pg_schemas": [self.pg_schema]} if self.pg_schema is not None else {}
    engine = _get_engine(self.db_conn, **kwargs)

src/dags/beheerkaart.py
```diff
@@ -137,4 +145,4 @@
 for table in create_tables:
     table >> swift_load_task
 
-swift_load_task >> provenance_renames >> swap_schema >> mkdir >> create_geopackage >> zip_geopackage >> upload_data >> grant_db_permissions  # noqa
+rename_cols >> swift_load_task >> provenance_renames >> swap_schema >> mkdir >> create_geopackage >> zip_geopackage >> upload_data >> grant_db_permissions  # noqa
```
jjmurre, line 148: Hier nog weer haakjes ivm leesbaarheid.

jjmurre (approved):
Ziet er goed uit.

PR #256, opened 2021-05-25
Take streaming response into account for HttpFetchOperator

jjmurre:
Turned out that for textual responses
the full text content of the response was being read into memory.

Introducing iter_lines to use memory more efficient.


ngaranko (approved):
LGTM

PR #262, opened 2021-05-31
Hotfix: reinstate imported SQL statements.

gkoller:
It looked they weren't used, but they were. Somewhere there
is code that does `globals()[select_statement]`. Added `noqa` statements
to prevent these import from being removed again in the future.

Fixed some MyPy and flake8 errors, while I was add it.

### DIFF
--- a/src/dags/basiskaart.py
+++ b/src/dags/basiskaart.py
@@ -1,26 +1,36 @@
 import logging
+from typing import Iterator
 
 from airflow import DAG
 from airflow.hooks.postgres_hook import PostgresHook
 from airflow.models import Variable
 from airflow.operators.postgres_operator import PostgresOperator
 from airflow.operators.python_operator import PythonOperator
-from sqlalchemy import create_engine
-from sqlalchemy.exc import SQLAlchemyError
-
-from common import (
-    env,
-    default_args,
-    slack_webhook_token,
-    DATAPUNT_ENVIRONMENT,
-    MessageOperator,
-)
+from common import DATAPUNT_ENVIRONMENT, MessageOperator, default_args, env, slack_webhook_token
 from dynamic_dagrun_operator import TriggerDynamicDagRunOperator
 from pgcomparator_cdc_operator import PgComparatorCDCOperator
 from postgres_permissions_operator import PostgresPermissionsOperator
 from provenance_rename_operator import ProvenanceRenameOperator
-from sql.basiskaart import CREATE_MVIEWS
-from sql.basiskaart import CREATE_TABLES
+from sql.basiskaart import CREATE_MVIEWS, CREATE_TABLES
+from sql.basiskaart import SELECT_GEBOUWVLAK_SQL as SELECT_GEBOUWVLAK_SQL  # noqa: F401
+from sql.basiskaart import (  # noqa: F401
+    SELECT_INRICHTINGSELEMENTLIJN_SQL as SELECT_INRICHTINGSELEMENTLIJN_SQL,
+)
+from sql.basiskaart import (  # noqa: F401
+    SELECT_INRICHTINGSELEMENTPUNT_SQL as SELECT_INRICHTINGSELEMENTPUNT_SQL,
+)
+from sql.basiskaart import (  # noqa: F401
+    SELECT_INRICHTINGSELEMENTVLAK_SQL as SELECT_INRICHTINGSELEMENTVLAK_SQL,
+)
+from sql.basiskaart import SELECT_LABELS_SQL as SELECT_LABELS_SQL  # noqa: F401
+from sql.basiskaart import SELECT_SPOORLIJN_SQL as SELECT_SPOORLIJN_SQL  # noqa: F401
+from sql.basiskaart import SELECT_TERREINDEELVLAK_SQL as SELECT_TERREINDEELVLAK_SQL  # noqa: F401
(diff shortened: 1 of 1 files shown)


src/dags/basiskaart.py
```diff
@@ -178,11 +194,22 @@ def create_basiskaart_dag(is_first, table_name, select_statement):
...
+        >> clean_up
+        >> trigger_next_dag
+        >> grant_db_permissions
+    )
 
     dag.doc_md = """
     #### DAG summary
-    This DAG contains BGT (basisregistratie grootschalige topografie) and KBK10 (kleinschalige basiskaart 10) and KBK50 (kleinschalige basiskaart 50) data
+    This DAG contains BGT (basisregistratie grootschalige topografie) and KBK10 (kleinschalige
+    basiskaart 10) and KBK50 (kleinschalige basiskaart 50) data
```
jjmurre, line 212: Very small nitpick. These are not really semantic linebreaks (term I love since you introduced it :-)

src/dags/basiskaart.py
```diff
@@ -193,7 +220,8 @@ def create_basiskaart_dag(is_first, table_name, select_statement):
     NA
     #### Prerequisites/Dependencies/Resourcing
     https://api.data.amsterdam.nl/v1/docs/datasets/basiskaart.html
-    Note: The basiskaart data is collected from the GOB objectstore and processed in the basiskaart DB => which is the source for this DAG.
+    Note: The basiskaart data is collected from the GOB objectstore and processed in the
```
jjmurre, line 223: Same here.

jjmurre (approved):
LGTM

src/dags/basiskaart.py
```diff
@@ -1,26 +1,36 @@
...
+from common import DATAPUNT_ENVIRONMENT, MessageOperator, default_args, env, slack_webhook_token
 from dynamic_dagrun_operator import TriggerDynamicDagRunOperator
 from pgcomparator_cdc_operator import PgComparatorCDCOperator
 from postgres_permissions_operator import PostgresPermissionsOperator
 from provenance_rename_operator import ProvenanceRenameOperator
-from sql.basiskaart import CREATE_MVIEWS
-from sql.basiskaart import CREATE_TABLES
+from sql.basiskaart import CREATE_MVIEWS, CREATE_TABLES
+from sql.basiskaart import SELECT_GEBOUWVLAK_SQL as SELECT_GEBOUWVLAK_SQL  # noqa: F401
+from sql.basiskaart import (  # noqa: F401
```
ngaranko, line 16: Could you please add explanation for noqa? Otherwise next person to change this file might remove imports again.

ngaranko (approved):

src/dags/basiskaart.py
gkoller: Good point. Will fix. I was a little to quick to satisfy flake8

src/dags/basiskaart.py
gkoller: I had missed your remark to document the `noqa`. 

Anyway, the why has already been documented in the Git commit message. I did it there as I always check the Git commit that introduced a change to find out why it was done. Can you imagine my frustration when people commit with explaining the why? Anyway I treat the commit history as part of the documentation of a project.

The `noqa: F401` already expresses the intention that they should be kept. I also don't feel like repeating myself. So when I have an atomic commit like this, I tend to not repeat myself in source code comments when I expect to  take the time to explain the why in the commit message.

Do you think I should do both? 

20210601_174225_screenshot

ngaranko: Most people never read comment messages, they just act. So it might be better to put comment on why this is needed, or yet better rewrite that `globals()[]` horror :) But that's story for another day.

gkoller: > Most people never read commit message

This make me sad. A proper and well written commit history can be so valuable. Code bases that have that tend to be joy to work with. Apart from the fact that I consider it documentation, it is also an indication how developers value their own work. Anyway, will add the comment as to why the `noqa: F401` is necessary in another PR.

> or yet better rewrite that globals()[]

Yeah, that's the ultimate solution. Though the issue with that is that the DAGs have virtually no tests. So I'm a little hesitant to modify code. And I often don't understand why things are done in a certain way, cause the _why_ is not explained in the commit messages that introduced the code that I want to change. 

So:

- no tests (I understand the why of that, it's hard in Airflow)
- no "why" in commit messages that introduced/modified certain pieces of code (we should really follow the practice describe in https://chris.beams.io/posts/git-commit/ more)
- wildly different code style to achieve apparently similar things
- Airflow is slow to run, so manually testing DAGs is time consuming; Eg iterating on the code is a slow process.

So it might take a while before I get around to rewriting that specific part. First auto generating DB tables based of schema defs instead of hardcoded SQL.

jjmurre: Rewriting `globals()[]` would be hard. Because that is connected to the way Airflow works (filling up the dagbag by reading the modules and only finding DAGs that are in the module namespace).

ngaranko: PS: I was not asking you to rewrite anything ;) Sorry for confusion.

As for commit messages: they are hard to rely on, as they are fragile: 
 - squash of branch before submitting PR (which is _sometimes_ considered good practice) will wipe whole history 
 - minor code refactoring (rename, line changes etc) will remove original reasoning for code introduction, not talking about situations when code is moved from one file to another
 - reading through history requires extra steps and may be really hard to navigate as codebase grows

Anyways, it's a good topic to pick up during standdown.

PR #263, opened 2021-05-31
Schedule openbareverlichting every 15 min.

jjmurre:
Also not using output_type="text" for HttpFetchOperator.
Fetching the data as-is, is much more efficient.

Some small flake8 + mypy fixes.

### DIFF
--- a/src/dags/openbareverlichting.py
+++ b/src/dags/openbareverlichting.py
@@ -38,16 +38,17 @@ tmp_dir = f"{SHARED_DIR}/{dag_id}"
 total_checks = []
 count_checks = []
 geo_checks = []
-check_name = {}
+
 
 # needed to put quotes on elements in geotypes for SQL_CHECK_GEO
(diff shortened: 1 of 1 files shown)


src/dags/openbareverlichting.py
```diff
@@ -38,16 +38,17 @@
 total_checks = []
 count_checks = []
 geo_checks = []
-check_name = {}
+
 
 # needed to put quotes on elements in geotypes for SQL_CHECK_GEO
-def quote(instr):
+def quote(instr: str) -> str:
```
gkoller, line 44: Not part of this PR, but any future SQL manipulation should be done using https://www.psycopg.org/docs/sql.html

gkoller (approved):

src/dags/openbareverlichting.py
jjmurre: 👍🏼

PR #265, opened 2021-06-01
Check for broken dags during deployment

jjmurre:
This python script is added to the initialization and startup of Airflow (in run.sh).
This makes the deployment fail, so broken dags are detected immediately during deployment.

### DIFF
--- a/src/scripts/checkdags.py
+++ b/src/scripts/checkdags.py
@@ -1,0 +1,21 @@
+"""Python script to check for broken DAGs.
+
+This scripts is running during the initialization and startup of Airflow during deployment.
+When DAGs are broken, this results in Airflow not being able to start.
+So, broken dags are detected during deployment.
+"""
+
+from pathlib import Path
+import subprocess
+
+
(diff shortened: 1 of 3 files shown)


ngaranko (approved):

src/scripts/checkdags.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+
+from pathlib import Path
+import subprocess
+
+
+dagsfolder = Path(__file__).parent / ".." / "dags"
+
+broken_dags = []
+for dagfile in dagsfolder.glob("*.py"):
```
gkoller, line 15: You should probably do `**/*.py`.

cvriel (approved):
Praktisch opgezet en nuttig!

src/scripts/checkdags.py
jjmurre: Good idea!

jjmurre: Hmm. Gives trouble (import errors), because of the special way Airflow handles PYTHON_PATH. So I leave it for now, only importing the dags at one level deep.

PR #266, opened 2021-06-01
Finance data processing

cvriel:
Data processing DAG for finance data for Tableau BI users.

### DIFF
--- a/src/dags/financien_grootboek.py
+++ b/src/dags/financien_grootboek.py
@@ -1,0 +1,132 @@
+from airflow import DAG
+from airflow.operators.postgres_operator import PostgresOperator
+from airflow.operators.python_operator import PythonOperator
+from importscripts.import_financien_grootboek import load_from_dwh
+from pgcomparator_cdc_operator import PgComparatorCDCOperator
+from postgres_check_operator import PostgresCheckOperator
+from postgres_permissions_operator import PostgresPermissionsOperator
+from provenance_rename_operator import ProvenanceRenameOperator
+from sqlalchemy_create_object_operator import SqlAlchemyCreateObjectOperator
+from sql.financien_grootboek import SQL_DROP_TMP_TABLE
+
+from common import (
+    default_args,
+    DATAPUNT_ENVIRONMENT,
+    slack_webhook_token,
+    MessageOperator,
+)
+from common.sql import SQL_CHECK_COUNT
+
+DATASTORE_TYPE: str = (
+    "acceptance" if DATAPUNT_ENVIRONMENT == "development" else DATAPUNT_ENVIRONMENT
+)
+
+data_schema_id: str = "financien"
+dag_id: str = "financien_grootboek"
+
+
+with DAG(
+    dag_id,
+    default_args={**default_args},
+    description="""Financiele gegevens uit de centrale administratie van het
+                    Amsterdams Financieel Systeem (AFS) beschikbaar gesteld via
+                    een datamart van DWH AMI (Amsterdamse Management Informatie).""",
+    schedule_interval="0 8 * * *",
+) as dag:
+
+    # 1. Post message on slack
+    slack_at_start = MessageOperator(
+        task_id="slack_at_start",
+        http_conn_id="slack",
+        webhook_token=slack_webhook_token,
+        message=f"Starting {dag_id} ({DATAPUNT_ENVIRONMENT})",
+        username="admin",
+    )
+
+    # 2. Load finance data into DB
+    load_dwh = PythonOperator(
+        task_id="load_from_ami_dwh",
+        python_callable=load_from_dwh,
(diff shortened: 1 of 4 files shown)


Bash- (approved):

src/dags/financien_grootboek.py
```diff
@@ -0,0 +1,132 @@
...
+
+from common import (
+    default_args,
+    DATAPUNT_ENVIRONMENT,
+    slack_webhook_token,
+    MessageOperator,
+)
+from common.sql import SQL_CHECK_COUNT
+
+DATASTORE_TYPE: str = (
```
gkoller, line 20: This exact definition is already present is 7 other DAGs:

- src/dags/anpr.py
- src/dags/beschermde_stadsdorpsgezichten.py
- src/dags/blackspots.py
- src/dags/beheerkaart.py
- src/dags/rioolnetwerk.py
- src/dags/varen.py
- src/dags/woningbouwplannen.py

I like to see this definition moved to the `common` module and removed from those 7 DAG and this new DAG.

src/dags/financien_grootboek.py
```diff
@@ -0,0 +1,132 @@
...
+    "acceptance" if DATAPUNT_ENVIRONMENT == "development" else DATAPUNT_ENVIRONMENT
+)
+
+data_schema_id: str = "financien"
+dag_id: str = "financien_grootboek"
+
+
+with DAG(
+    dag_id,
+    default_args={**default_args},
```
gkoller, line 30: You seem to be destructing the dictionary only to create another one. You would do that if you expect the DAG to modify the `default_args`. However it does not. On the contrary, it does a `deepcopy` which is safer than the destructuring using `**`. Hence you should change this to: `default_args=default_args`

Note: see `DAG.__init__(...)`, it does `self.default_args = copy.deepcopy(default_args or {})`

src/dags/financien_grootboek.py
```diff
@@ -0,0 +1,132 @@
...
+        sql=SQL_DROP_TMP_TABLE,
+        params=dict(tablename=f"{dag_id}_new"),
+    )
+
+    # 8. Grant database permissions
+    grant_db_permissions = PostgresPermissionsOperator(task_id="grants", dag_name=data_schema_id)
+
+
+# FLOW
+[
```
gkoller, line 104: No need to create a one element list here. The parenthesis used in other DAG do not create a one element iterable either (a tuple). They are just there for structuring/formatting purposes)

src/dags/importscripts/import_financien_grootboek.py
```diff
@@ -0,0 +1,138 @@
...
+            ,org.RVE_NAAM as "rve_naam"
+            ,org.CLUSTER_CODE as "cluster_code"
+            ,org.CLUSTER_NAAM as "cluster_naam"
+            FROM DATAHUB.FI2_DHB_FCT_GROOTBOEK GB
+            INNER JOIN DATAHUB.FI2_DHB_DIM_RELATIE REL ON REL.ID = GB.REL_ID
+            INNER JOIN DATAHUB.FI2_DHB_DIM_ORGANISATIE org ON ORG.ID = GB.ORG_ID
+            INNER JOIN DATAHUB.FI2_DHB_DIM_GROOTBOEKREKENING gbr ON gbr.ID = GB.gbr_ID
+            INNER JOIN DATAHUB.FI2_DHB_DIM_VERPLICHTINGREGEL vpr ON vpr.ID = GB.VPR_ID
+            INNER JOIN DATAHUB.FI2_DHB_DIM_WERKORDER wor ON wor.ID = GB.WOR_ID
+            WHERE 1=1
```
gkoller, line 75: I see this often in your SQL statements. It does nothing. Might I suggest you remove it?

src/dags/importscripts/import_financien_grootboek.py
```diff
@@ -0,0 +1,138 @@
...
+            "rve_code": Text(),
+            "rve_naam": Text(),
+            "cluster_code": Text(),
+            "cluster_naam": Text(),
+        }
+        df.to_sql(
+            table_name, db_engine, if_exists="replace", index_label="id", index=True, dtype=dtype
+        )
+        with db_engine.connect() as connection:
+            connection.execute(f"ALTER TABLE {table_name} ADD PRIMARY KEY (ID)")
```
gkoller, line 138: We should not be using regular string formatting for constructing SQL queries. It's probably safe here, but the practice gets copied over and over and as such persists. Better to use the features of `psycopg2`: https://www.psycopg.org/docs/sql.html

Using that the statement would become (untested):
```python
from psycopg2 import sql
...
connection.execute(
    sql.SQL("ALTER TABLE {table_name} ADD PRIMARY KEY (ID)")
        .format(table_name=sql.Identifier(table_name)
)
```

PR #268, opened 2021-06-02
Moved `set -e` closer to execution scripts/checkdags.py

cvriel:
The set -e configuration in run.sh is used to stop Airflow when one of the DAG's cannot be executed due to some error. An extra safety check before deployment of changes.

In the run.sh one of the cmd's is to delete and create Airflow connections. When runned locally (for the first time) the connection are not yet created when the cmd to delete the connection is executed (after that it will create the connection).

Because the connection does not yet exists when deleted, it will give an exit code 1. Causing the Airflow to stop running.

As a quick fic the `set -e` is set just befor e the scripts/checkdags.py. So it will be active for the check on correct DAG.

### DIFF
--- a/src/scripts/run.sh
+++ b/src/scripts/run.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e # stop on errors
-
 export AIRFLOW__CORE__SQL_ALCHEMY_CONN=${AIRFLOW__CORE__SQL_ALCHEMY_CONN:-`echo $AIRFLOW_CONN_POSTGRES_DEFAULT | cut -d'?' -f 1`}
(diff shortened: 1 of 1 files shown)


c-arnarez (approved):

src/scripts/run.sh
```diff
@@ -94,7 +92,7 @@ airflow variables import vars/vars.json
 # to run the checkscript.
 # Make sure that the the containing shell script (run.sh)
 # stops on errors (set -e).
```
jjmurre, line 94: Het set -e kan nog uit het commentaar worden verwijderd.

jjmurre (approved):
LGTM

PR #271, opened 2021-06-03
AB#4367 notify source owners failed import

gkoller:
This PR introduce an `on_error_callback` that, when a DAG run fails, can retrieve the `contactPoint` from the associated Amsterdam Schema to find out who to email a failure notification to.

It also introduces tighter pre-commit checks for this project.

When reviewing please pay special attention to commit fa09e55a0bed53b5e553564a6f682c9ef4f2534c The commit explains the why it needs some special care (TL;DR there is no always a direct mapping between `dag_id` and `dataset_id`.)

As usual, probably best reviewed per commit

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,6 +1,7 @@
 .PHONY: install
 install:
 	pip install -r requirements_dev.txt
+	pre-commit clean
 	pre-commit install
 
 .PHONY: sync
--- a/src/plugins/contact_point/models.py
+++ b/src/plugins/contact_point/models.py
@@ -1,0 +1,29 @@
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class ContactPoint:
+    """Represent an Amsterdam Schema ``contactPoint``.
+
+    The Amsterdam Schema defines the ``contactPoint``, but also its attributes, to be entirely
+    optional.
+    """
+
+    name: Optional[str] = None
+    email: Optional[str] = None
+
+    def __str__(self) -> str:
+        """Return human readable string representation.
+
+        We strive for ``"name <email>"``, but it can even be just
+         - an empty string,
+         - only the name
+         - only the email address.
+
+        Returns:
+            Human readable string representation.
+        """
+        name = self.name if self.name else ""
+        email = f"<{self.email}>" if self.email else ""
+        return f"{name} {email}".strip()
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,5 @@ dmypy.json
 
 # IntelliJ based IDEs
 .idea/
+
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -144,4 +144,15 @@ dmypy.json
 # IntelliJ based IDEs
 .idea/
 
+# Visual Studio Code IDE
 .vscode/
+
+# Real config lives in src/config/airflow.cfg
+# ignored path is symlink to allow it run locally with
+# export AIRFLOW_HOME=$(pwd)/src
+src/airflow.cfg
+
+# When running locally a webserver_config
+# will be created with default values.
+# This too should not be checked in.
+src/webserver_config.py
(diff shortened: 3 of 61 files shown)


.flake8
```diff
@@ -1,25 +1,53 @@
...
-  # Allow raise Exception()
-  R102,
-exclude=.git,
-        **/migrations/*,
-        .env,
-        docs,
-        scripts,
-format = ${cyan}%(path)s${reset}:${yellow_bold}%(row)d${reset}:${green_bold}%(col)d${reset}: ${red_bold}%(code)s${reset} %(text)s
-max-complexity=8
+# A = builtins
```
ngaranko, line 2: Kan dit naar aparte PR getrekt worden samen met alle andere precommit tweaks?

src/plugins/contact_point/models.py
```diff
@@ -0,0 +1,29 @@
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class ContactPoint:
+    """Represent an Amsterdam Schema ``contactPoint``.
+
+    The Amsterdam Scheme defines the ``contactPoint``, but also its attributes, to be entirely
```
lbam, line 9: Scheme => Schema

.flake8
gkoller: Yes, I could do that. But why if I may ask so? I tightened the checks so that the new code I would write as part of this PR would be fine.

src/plugins/contact_point/models.py
gkoller: Thanks. Fixed.

.flake8
ngaranko: For clarity. 
Just like with commit messages, separation of logic introduced into repository should be logical.
Code checks have absolutely nothing to do with notifications, but as they are introduced in same PR - some may get confused and think that they are related and notifications code may need tighter checks.

Obviously ^ will never happen, but it's not a reason to put unrelated changes into single PR.

src/Makefile
```diff
@@ -15,4 +16,4 @@ requirements: requirements.in
 upgrade:
 	pur --patch=Django -r requirements.in
 	pur --no-recursive -r requirements_dev.txt
-	pip-compile -v --upgrade --output-file requirements.txt requirements.in
```
ngaranko, line 18: This means we're not doing non-pinned package upgrades inside `make upgrade` anymore?

ngaranko (approved):

.flake8
gkoller: Ah, I get it now. Thank for the explanation.

Personally I am okay with PR's that include changes that are tangentially related to the topic of the PR (in this case I wanted my new code to adhere to the strictest style checks). As long as the changes are introduced in separate commits (with clear commit messages :-)). After all the commits are what is normally retained after the PR has been_rebased and merged_. After all the commits are part of the Git history, the PR is not (its part of the Git{Hub,Lab} history. So after a _rebase and merge_ you can't see whether the changes were introduced by one or two PRs.

However for reviewing purpose it might make sense to split them up. Hence I'll do so.

We should definitely discuss this more. Small PRs and _squash and merge_. Larger PRs and _rebase and merge_.

src/Makefile
gkoller: Hmm, good point. I think this change was a mistake. I initially thought that it would disregard the `--patch=Django` argument to `pur`. Eg `pur` would only patch-upgrade Django, but then the `--upgrade` argument to `pip-compile` would upgrade it to another major version (if available).

Anyway tested it out and it seems that `pip-compile` treats `requirements.in` and a constraint in the presence of `--upgrade`. So it is safe after all. Will revert.

.flake8
gkoller: Done, moved the tighter `flake8` and `pre-commit` check to a new PR: https://github.com/Amsterdam/dataservices-airflow/pull/272

ngaranko (approved):

PR #273, opened 2021-06-04
Overall adjustments

cvriel:
- constant DATASTORE_TYPE moved to common
- replaced all quote methods in DAG's with common quote_string
- used psycopg2 in SQL statement financien_grootboek
- make return connection in common

### DIFF
--- a/src/dags/common/db.py
+++ b/src/dags/common/db.py
@@ -19,8 +19,9 @@ class DatabaseEngine:
         self.db = self.connection.schema
 
 
-def get_connection(postgres_conn_id: str = "postgres_default") -> Engine:
-    """Return database connection."""
+def get_postgreshook_instance(postgres_conn_id: str = "postgres_default") -> PostgresHook:
+    """Return a postgreshook instance,
+    so it can be used to get connection i.e."""
     connection = PostgresHook(postgres_conn_id=postgres_conn_id)
     return connection
 
--- a/src/dags/common/db.py
+++ b/src/dags/common/db.py
@@ -1,10 +1,10 @@
 import dsnparse
 from airflow import AirflowException
 from airflow.hooks.oracle_hook import OracleHook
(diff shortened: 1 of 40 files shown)


src/dags/common/db.py
```diff
@@ -19,6 +19,12 @@ def __init__(self, postgres_conn_id="postgres_default"):
         self.db = self.connection.schema
 
 
+def get_connection(postgres_conn_id: str = "postgres_default") -> Engine:
```
gkoller, line 22: This type annotation does not seem correct. `Engine` is not in the type hierarchy of `PostgresHook`

20210607_083135_screenshot

src/dags/common/db.py
```diff
@@ -19,6 +19,12 @@ def __init__(self, postgres_conn_id="postgres_default"):
...
+    """Return database connection."""
+    connection = PostgresHook(postgres_conn_id=postgres_conn_id)
```
gkoller, line 24: Naming the `PostgresHook` `connection` is kind of confusing. `PostgresHook` is not a connection, it encapsulates one, namely `psycopg2._psycopg.connection.connection` to be retrieved using `PostgresHook`'s `get_conn` method.

What I'm saying that the message that this function conveys is rather ambiguous:
- the function is called `get_connection`
- that does not *seem* to return a connection, but an `Engine`
- though the implementation, although the instance is named `connection`, in reality returns a `PostgresHook` instance.

gkoller (approved):
Nice clean up. One remaining issue re: `common.db.get_connection`

src/dags/common/db.py
cvriel: Oeps, pas ik aan

src/dags/common/db.py
cvriel: You are right, the get_connection is confusing and does not fit. I will change it's name to be more appropriate

PR #276, opened 2021-06-08
excl log_cleanup dag to perm operator

cvriel:
The permissions operator (used in the permissions DAG) crashes at some time. 

This operator uses the Amsterdam schema definition to retrieve the database objects (tables, scopes) to setup the database permissions. It uses the DAG names to retrieve the corresponding schema. However, the log_cleanup DAG does not have a schema definition. And therefore must be excluded from the logic.


jjmurre (approved):
LGTM

PR #278, opened 2021-06-10
AB#8204 New source location

cvriel:
The source files OOV overlastgebieden are moved to a different location. To succesfully connect to the location a SFTPOperator was used. This operator is a out-of-the-box operator in Airflow. In addition extra libraries where needed to execute the operator.

Further more:
- The ogr2ogr operator replaced the older bash cmd (doing the same).
- Some mypy, flake8 corrections where made that weren't addressed yet.

### DIFF
--- a/src/dags/overlastgebieden.py
+++ b/src/dags/overlastgebieden.py
@@ -1,53 +1,50 @@
 import operator
+from typing import Dict
+
 from airflow import DAG
 from airflow.models import Variable
 from airflow.operators.bash_operator import BashOperator
-from airflow.operators.postgres_operator import PostgresOperator
 from airflow.operators.dummy_operator import DummyOperator
-
-from contact_point.callbacks import get_contact_point_on_failure_callback
-from provenance_rename_operator import ProvenanceRenameOperator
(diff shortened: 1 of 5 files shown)


src/dags/overlastgebieden.py
```diff
@@ -1,53 +1,50 @@
...
-    GEO_CHECK,
-)
-
-dag_id = "overlastgebieden"
+dag_id: str = "overlastgebieden"
 
-variables_overlastgebieden = Variable.get("overlastgebieden", deserialize_json=True)
-files_to_download = variables_overlastgebieden["files_to_download"]
-tables_to_create = variables_overlastgebieden["tables_to_create"]
+variables_overlastgebieden: Dict = Variable.get("overlastgebieden", deserialize_json=True)
```
jjmurre, line 28: Voor de `Dict` en `List` types nog de datatypes van de inhoud, of de key/value tovoegen.

jjmurre (approved):
👍🏼

src/dags/overlastgebieden.py
cvriel: :+1:

PR #279, opened 2021-06-14
AB#3811 stroomstoringen regio Amsterdam

cvriel:
A new data processing logic for electra blackouts in region Amsterdam. The data is maintained by Liander.

### DIFF
--- a/src/dags/sql/stroomstoringen.py
+++ b/src/dags/sql/stroomstoringen.py
@@ -1,0 +1,71 @@
+# The data contains epoch times.
+# In order to translate it to a timestamp, the data is updated.
+CONVERT_DATE_TIME = """
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_gemeld TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_schatting TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_eind TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN date_end TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_change TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_service_update TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN log_date TYPE varchar(50);
+
+UPDATE {{ params.tablename }}
+SET storing_datum_gemeld = to_timestamp(rpad(storing_datum_gemeld, 13, '0')::bigint / 1000),
+    storing_datum_schatting = to_timestamp(rpad(storing_datum_schatting, 13, '0')::bigint / 1000),
+    storing_datum_eind = to_timestamp(rpad(storing_datum_eind, 13, '0')::bigint / 1000),
+    date_end = to_timestamp(rpad(date_end, 13, '0')::bigint / 1000),
+    storing_datum_change = to_timestamp(rpad(storing_datum_change, 13, '0')::bigint / 1000),
+    storing_service_update = to_timestamp(rpad(storing_service_update, 13, '0')::bigint / 1000),
+    log_date = to_timestamp(rpad(log_date, 13, '0')::bigint / 1000)
+WHERE 1=1;
+COMMIT;
+
+ALTER TABLE {{ params.tablename }} ADD COLUMN storing_datum_gemeld_tmp TIMESTAMP
+    without time zone NULL;
+UPDATE {{ params.tablename }} SET storing_datum_gemeld_tmp = storing_datum_gemeld::TIMESTAMP;
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_gemeld TYPE TIMESTAMP
+    without time zone USING storing_datum_gemeld_tmp;
+ALTER TABLE {{ params.tablename }} DROP COLUMN storing_datum_gemeld_tmp;
+
+ALTER TABLE {{ params.tablename }} ADD COLUMN storing_datum_schatting_tmp TIMESTAMP
+    without time zone NULL;
+UPDATE {{ params.tablename }} SET storing_datum_schatting_tmp = storing_datum_schatting::TIMESTAMP;
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_schatting TYPE TIMESTAMP
+    without time zone USING storing_datum_schatting_tmp;
+ALTER TABLE {{ params.tablename }} DROP COLUMN storing_datum_schatting_tmp;
+
+ALTER TABLE {{ params.tablename }} ADD COLUMN storing_datum_eind_tmp TIMESTAMP
+    without time zone NULL;
+UPDATE {{ params.tablename }} SET storing_datum_eind_tmp = storing_datum_eind::TIMESTAMP;
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_eind TYPE TIMESTAMP
+    without time zone USING storing_datum_eind_tmp;
+ALTER TABLE {{ params.tablename }} DROP COLUMN storing_datum_eind_tmp;
+
+ALTER TABLE {{ params.tablename }} ADD COLUMN date_end_tmp TIMESTAMP
+    without time zone NULL;
+UPDATE {{ params.tablename }} SET date_end_tmp = date_end::TIMESTAMP;
+ALTER TABLE {{ params.tablename }} ALTER COLUMN date_end TYPE TIMESTAMP
+    without time zone USING date_end_tmp;
+ALTER TABLE {{ params.tablename }} DROP COLUMN date_end_tmp;
+
+ALTER TABLE {{ params.tablename }} ADD COLUMN storing_datum_change_tmp TIMESTAMP
+    without time zone NULL;
+UPDATE {{ params.tablename }} SET storing_datum_change_tmp = storing_datum_change::TIMESTAMP;
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_change TYPE TIMESTAMP
+    without time zone USING storing_datum_change_tmp;
+ALTER TABLE {{ params.tablename }} DROP COLUMN storing_datum_change_tmp;
+
+ALTER TABLE {{ params.tablename }} ADD COLUMN storing_service_update_tmp TIMESTAMP
+    without time zone NULL;
+UPDATE {{ params.tablename }} SET storing_service_update_tmp = storing_service_update::TIMESTAMP;
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_service_update TYPE TIMESTAMP
+    without time zone USING storing_service_update_tmp;
+ALTER TABLE {{ params.tablename }} DROP COLUMN storing_service_update_tmp;
+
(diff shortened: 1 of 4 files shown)


src/dags/stroomstoringen.py
```diff
@@ -0,0 +1,137 @@
...
+from postgres_rename_operator import PostgresTableRenameOperator
+from provenance_rename_operator import ProvenanceRenameOperator
+from sql.stroomstoringen import CONVERT_DATE_TIME
+
+dag_id: str = "stroomstoringen"
+variables: dict = Variable.get("stroomstoringen", deserialize_json=True)
+endpoint_url: str = variables["data_endpoints"]["stroomstoringen"]
+
+
+today = datetime.now().strftime("%m-%d-%Y")
```
gkoller, line 29: You should probably take the timezone into account. Even when you are only after the date, you don't know before hand when a DAG is going to be scheduled. You will get an incorrect result if the DAG is scheduled during an daylight savings change. And even if you do know it will not be issue in this case, within AIrflow we generally do take timezones into account. Given the fact that code gets copied it is best to do it properly.

Airflow uses the `pendulum` library. I'm not sure if pendulum timezones are compatible with those of the Python stdlib. Hence it is best to use `pendulum` exclusively

```python
import pendulum
from airflow.settings import TIMEZONE

# uppercase as it is used as a constant
TODAY = pendulum.now(TIMEZONE).format("MM-DD-YYYY")
```

BTW is the date format really month, day, year in that order?

src/dags/sql/stroomstoringen.py
```diff
@@ -0,0 +1,71 @@
...
+
+UPDATE {{ params.tablename }}
+SET storing_datum_gemeld = to_timestamp(rpad(storing_datum_gemeld, 13, '0')::bigint / 1000),
+    storing_datum_schatting = to_timestamp(rpad(storing_datum_schatting, 13, '0')::bigint / 1000),
+    storing_datum_eind = to_timestamp(rpad(storing_datum_eind, 13, '0')::bigint / 1000),
+    date_end = to_timestamp(rpad(date_end, 13, '0')::bigint / 1000),
+    storing_datum_change = to_timestamp(rpad(storing_datum_change, 13, '0')::bigint / 1000),
+    storing_service_update = to_timestamp(rpad(storing_service_update, 13, '0')::bigint / 1000),
+    log_date = to_timestamp(rpad(log_date, 13, '0')::bigint / 1000)
+WHERE 1=1;
```
gkoller, line 20: `1=1` does not do anyhing

src/dags/sql/stroomstoringen.py
```diff
@@ -0,0 +1,71 @@
...
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_gemeld TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_schatting TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_eind TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN date_end TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_datum_change TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN storing_service_update TYPE varchar(50);
+ALTER TABLE {{ params.tablename }} ALTER COLUMN log_date TYPE varchar(50);
+
+UPDATE {{ params.tablename }}
+SET storing_datum_gemeld = to_timestamp(rpad(storing_datum_gemeld, 13, '0')::bigint / 1000),
```
gkoller, line 13: Your comment says: `# The data contains epoch times.` if that is the case I think `to_timestamp(<column_name>)` is sufficient?

If it isn't sufficient could you then explain why you are using `rpad` and `/ 1000`?

src/dags/sql/stroomstoringen.py
```diff
@@ -0,0 +1,71 @@
...
+    storing_datum_eind = to_timestamp(rpad(storing_datum_eind, 13, '0')::bigint / 1000),
+    date_end = to_timestamp(rpad(date_end, 13, '0')::bigint / 1000),
+    storing_datum_change = to_timestamp(rpad(storing_datum_change, 13, '0')::bigint / 1000),
+    storing_service_update = to_timestamp(rpad(storing_service_update, 13, '0')::bigint / 1000),
+    log_date = to_timestamp(rpad(log_date, 13, '0')::bigint / 1000)
+WHERE 1=1;
+COMMIT;
+
+ALTER TABLE {{ params.tablename }} ADD COLUMN storing_datum_gemeld_tmp TIMESTAMP
+    without time zone NULL;
```
gkoller, line 24: Why are you storing timestamps without a timezone?  How do external systems know what timezone the timestamps are in?

src/dags/sql/stroomstoringen.py
```diff
@@ -0,0 +1,71 @@
...
+UPDATE {{ params.tablename }} SET storing_datum_gemeld_tmp = storing_datum_gemeld::TIMESTAMP;
```
gkoller, line 25: If you allow these columns to have a timezone you don't need an explicit cast to `::TIMEZONE`

src/dags/sql/stroomstoringen.py
```diff
@@ -0,0 +1,71 @@
...
+    without time zone USING storing_service_update_tmp;
+ALTER TABLE {{ params.tablename }} DROP COLUMN storing_service_update_tmp;
+
+ALTER TABLE {{ params.tablename }} ADD COLUMN log_date_tmp TIMESTAMP
+    without time zone NULL;
+UPDATE {{ params.tablename }} SET log_date_tmp = log_date::TIMESTAMP;
+ALTER TABLE {{ params.tablename }} ALTER COLUMN log_date TYPE TIMESTAMP
+    without time zone USING log_date_tmp;
+ALTER TABLE {{ params.tablename }} DROP COLUMN log_date_tmp;
+"""
```
gkoller, line 71: This all seems very complex just to alter the data type of a column. I looked at the [PostgreSQL documentation ](https://www.postgresql.org/docs/12/sql-altertable.html ) and there is a better way:

Untested:
```psql
ALTER TABLE {{ params.tablename }}
    ALTER COLUMN storing_datum_gemeld SET DATA TYPE timestamp with timezone using to_timestamp
    ALTER COLUMN storing_datum_schatting SET DATA TYPE timestamp with timezone using to_timestamp
    ALTER COLUMN storing_datum_eind SET DATA TYPE timestamp with timezone using to_timestamp
    ALTER COLUMN date_end SET DATA TYPE timestamp with timezone using to_timestamp
    ALTER COLUMN storing_datum_change SET DATA TYPE timestamp with timezone using to_timestamp
    ALTER COLUMN storing_service_update SET DATA TYPE timestamp with timezone using to_timestamp
    ALTER COLUMN log_date SET DATA TYPE timestamp with timezone using to_timestamp
```

src/dags/stroomstoringen.py
```diff
@@ -0,0 +1,137 @@
...
+    convert_datetime = PostgresOperator(
+        task_id="convert_datetime",
+        sql=CONVERT_DATE_TIME,
+        params={"tablename": f"{dag_id}_{dag_id}_new"},
+    )
+
+    # 6. Rename COLUMNS based on Provenance
+    provenance_translation = ProvenanceRenameOperator(
+        task_id="rename_columns",
+        dataset_name=f"{dag_id}",
```
gkoller, line 92: You are not formatting anything here. Simply use `dataset_name={dag_id},`

src/dags/stroomstoringen.py
```diff
@@ -0,0 +1,137 @@
...
+from sql.stroomstoringen import CONVERT_DATE_TIME
+
+dag_id: str = "stroomstoringen"
+variables: dict = Variable.get("stroomstoringen", deserialize_json=True)
+endpoint_url: str = variables["data_endpoints"]["stroomstoringen"]
+
+
+today = datetime.now().strftime("%m-%d-%Y")
+
+tmp_dir: str = f"{SHARED_DIR}/{dag_id}"
```
gkoller, line 31: We are half way between using `Path` versus string. For example the `HttpFetchOperator` already does accept a `Path` object for `tmp_file`. But the `Ogr2OgrOperator` does not. Personally I'd rather see us dealing with `Path` objects for as long as possible and only convert to a string when needed.

But as not all operators support it yet, a string should be okay for now. Though this could be an opportunity to convert the `Ogr2OgrOperator` to accept both `Path` and `str` similarly to `HttpFetchOperator`

src/dags/stroomstoringen.py
```diff
@@ -0,0 +1,137 @@
...
+total_checks: list = []
+count_checks: list = []
+geo_checks: list = []
+check_name: dict = {}
```
gkoller, line 35: Love it that you are using type annotations so faithfully. But in these cases MyPy and other tools will automatically infer that these are `list`s and `dict`s. What these tools can't infer is what kind of `list`s and `dict`s these are.  Providing that, provides meaningful information to MyPy.

Eg `List[str]` or `List[int]` or in case of `dict`s `Dict[str, int]` (just using something random here, don't know what it really should be)

src/dags/stroomstoringen.py
```diff
@@ -0,0 +1,137 @@
...
+db_conn: object = DatabaseEngine()
```
gkoller, line 36: This is correct, but also an extreme generalization (almost everything is an `object`).
This is better:
```python
db_conn: DatabaseEngine = DatabaseEngine()
```

Though:
```python
db_conn  = DatabaseEngine()
```
works equally well as MyPy can infer the type from the assignment.

src/dags/stroomstoringen.py
gkoller: BTW where are these global variables used?

src/dags/stroomstoringen.py
cvriel: :+1: excellent 
Yes, its month, day, year (in that order). An American format apparently.

src/dags/sql/stroomstoringen.py
cvriel: Oeps, that is deeply rooted in my system I quess...

src/dags/sql/stroomstoringen.py
cvriel: Good one, the dates are stored in millseconds. Unix timestamps measures time with seconds, and not milliseconds (in Postgres too). Therefore the values must be devided by 1000 before converting to timestamp. I wil add it to the comment :+1:

src/dags/sql/stroomstoringen.py
cvriel: Yes, agree, will change it to with timezone

src/dags/sql/stroomstoringen.py
cvriel: :+1:

src/dags/sql/stroomstoringen.py
cvriel: Thanks :+1:

src/dags/stroomstoringen.py
cvriel: Oeps...

src/dags/stroomstoringen.py
cvriel: Oeps...never used them, unfortunately it passed through the checkout. Strange, I will have a look at it...

src/dags/stroomstoringen.py
cvriel: Ah didn't know that!

PR #282, opened 2021-06-18
Remove old incomplete way of updating DAGs.

gkoller:
This DAG updating DAG only updated the DAGs not taking into account:
- variables
- plugins
- Airflow restarts (plugins are only loaded once)

Hence it could never have worked correctly in all use-cases.


cvriel (approved):

jjmurre (approved):
LGTM, like we discussed, this DAG is never used and has too much potential to shoot yourself in the foot.

PR #283, opened 2021-06-20
Added use of the Path class in the ogr2ogr_operator

cvriel:
To be consistent with the http_fetch_operator, the parameters for file in- and output in the org2ogr operator is using a Path instance for the argument value (string full path to file). It can be used to create the desired directories if not exists yet.


jjmurre (approved):
LGTM

PR #285, opened 2021-06-22
Resource bouwstroompunten extra fields

cvriel:
The bouwstroompunten has some extra fields in the output. In addition, it opens the opportunity to update the dag so it uses the latest operator solutions. Making it more consistent with other dags.

### DIFF
--- a/src/dags/bouwstroompunten.py
+++ b/src/dags/bouwstroompunten.py
@@ -193,7 +193,7 @@ with DAG(
 
 dag.doc_md = """
     #### DAG summary
-    This DAG contains power stations data for event or mar
+    This DAG contains power stations data for event or market
     #### Mission Critical
     Classified as 2 (beschikbaarheid [range: 1,2,3])
     #### On Failure Actions
--- a/src/dags/bouwstroompunten.py
+++ b/src/dags/bouwstroompunten.py
@@ -1,7 +1,7 @@
 import json
 import operator
 from pathlib import Path
-from typing import Dict
+from typing import Any, Union
 
 import dsnparse
 import requests
@@ -45,7 +45,7 @@ geo_checks = []
 db_conn: object = DatabaseEngine()
 
 
-def get_token() -> Dict[str, str]:
+def get_token() -> Union[str, Any]:
     """Getting the access token for calling the data endpoint.
 
     Returns:
@@ -61,12 +61,18 @@ def get_token() -> Dict[str, str]:
     token_url = f"{base_endpoint}/{auth_endpoint}"
     token_payload = {"identifier": user, "password": password}
     token_headers = {"content-type": "application/json"}
-    token_request = requests.post(
-        token_url,
-        data=json.dumps(token_payload),
-        headers=token_headers,
-        verify=True,
-    )
+
+    try:
+        token_request = requests.post(
+            token_url,
+            data=json.dumps(token_payload),
+            headers=token_headers,
+            verify=True,
+        )
+        token_request.raise_for_status()
+    except requests.exceptions.HTTPError as err:
+        raise requests.exceptions.HTTPError("Something went wrong", err.response.text)
+
     token_load = json.loads(token_request.text)
     return token_load["jwt"]
 
--- a/src/dags/bouwstroompunten.py
+++ b/src/dags/bouwstroompunten.py
@@ -1,23 +1,26 @@
 import json
 import operator
+from typing import Any
 
+import dsnparse
 import requests
 from airflow import DAG
 from airflow.models import Variable
-from airflow.operators.bash import BashOperator
-from airflow.operators.python import PythonOperator
-from airflow.providers.postgres.operators.postgres import PostgresOperator
+from airflow.operators.bash_operator import BashOperator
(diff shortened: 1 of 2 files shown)


src/dags/bouwstroompunten.py
```diff
@@ -1,23 +1,26 @@
...
 
+import dsnparse
 import requests
 from airflow import DAG
 from airflow.models import Variable
-from airflow.operators.bash import BashOperator
-from airflow.operators.python import PythonOperator
-from airflow.providers.postgres.operators.postgres import PostgresOperator
+from airflow.operators.bash_operator import BashOperator
+from airflow.operators.postgres_operator import PostgresOperator
```
gkoller, line 10: The original imports were good. This changes it to deprecated ones.
You should use:
```
from airflow.operators.python import PythonOperator
from airflow.providers.postgres.operators.postgres import PostgresOperator
```

src/dags/bouwstroompunten.py
```diff
@@ -89,89 +91,98 @@ def get_data():
...
-    # 10. RUN bundled CHECKS
+    # 9. RUN bundled CHECKS
     multi_checks = PostgresMultiCheckOperator(task_id="multi_check", checks=total_checks)
 
-    # 11. Grant database permissions
+    # 10. Grant database permissions
     grant_db_permissions = PostgresPermissionsOperator(task_id="grants", dag_name=dag_id)
 
 
 (
```
gkoller, line 181: This should be indented one level more. The workflow specification (how the tasks are linked together) should be part of the `with` statement.

src/dags/bouwstroompunten.py
```diff
@@ -38,34 +41,33 @@
...
 
+db_conn: object = DatabaseEngine()
 
-# data connection
-def get_data():
-    """getting the the access token and calling the data endpoint"""
 
-    # get token
-    token_url = f"{base_url}/{auth_endpoint}"
+def get_token() -> Any:
```
gkoller, line 47: By returning `Any` you are basically turning type annotations off. If possible you should be more specific.

You seem to be doing `return token_load["jwt"]`. What is that. Is it:
* a Base64Url encode string? If so, annotated the function return type with `-> str`
* a decoded JWT token? If so, annotate the function return type with `-> Dict[str, str] (assuming it's a dict, don't know much about JWT)

src/dags/bouwstroompunten.py
```diff
@@ -89,89 +91,98 @@ def get_data():
     mkdir = BashOperator(task_id="mkdir", bash_command=f"mkdir -p {tmp_dir}")
```
gkoller, line 91: If you are using more modern operators might I suggest you you the `mk_dir` task here?

```python
from common.path import mk_dir
...
mkdir = mk_dir(tmp_dir, clean_if_exists=True)
```
Not sure if you want it cleaned though. If not, leave the argument out (default value is False), or set it explicitly to false.

The task `mk_dir` is our first usage of the Airflow 2.0 Flow based API. Much more convenient API than the Operator based one. See also the `anpr` DAG that already uses `mk_dir`

src/dags/bouwstroompunten.py
gkoller: I suspect is should be `-> str` not that I have seen how it is used in:

```python
...
"Authorization": f"Bearer {get_token()}",
...
```

src/dags/bouwstroompunten.py
```diff
@@ -89,89 +91,98 @@ def get_data():
...
+        mode="PostgreSQL",
+        db_conn=db_conn,
     )
 
-    # 6. Drop Exisiting TABLE
+    # 5. Drop Exisiting TABLE
     drop_tables = PostgresOperator(
         task_id="drop_existing_table",
         sql=[
             f"DROP TABLE IF EXISTS {dag_id}_{dag_id} CASCADE",
```
gkoller, line 123: I know this is existing code, but as you are touching the DAG anyway ...

We should never be using regular string formatting for creating SQL strings. Fortunately we are not processing user input here. Hence the potential for SQL injection is extremely low. But code gets copied around or mimicked hence it should be done properly.

The PostgresqlOperator has its own SQL string formatting capabilities. So it is best to use those:

```python
    drop_tables = PostgresOperator(
        task_id="drop_existing_table",
        sql=[
            "DROP TABLE IF EXISTS {{ params.table_id }} CASCADE",
        ],
        params={"table_id": f"{dag_id}_{dag_id}"} 
    )
```

gkoller (approved):
Couple of review to be addressed, but otherwise good to go :+1:

src/dags/bouwstroompunten.py
cvriel: :+1:

src/dags/bouwstroompunten.py
cvriel: :+1:

src/dags/bouwstroompunten.py
cvriel: Nice addition, didn't know that!

src/dags/bouwstroompunten.py
cvriel: Ok check!

PR #287, opened 2021-06-23
Escape literal name when compiling as regular expression

lbam:
Spotted this while checking out  #286. We accept a literal name (string) or a regular expression for `SqlAlchemyCreateObjectOperator.data_table_name`. I suppose a valid name is also a valid regexp, but it's easier to catch bugs when it's escaped first.


jjmurre (approved):
Good idea, better safe than sorry.

PR #288, opened 2021-06-24
Guido/remove old vsd jobs

gkoller:
* Remove unused `# type: ignore` for `@apply_defaults`
* Rename identically named non-package modules in src tree
* Delete old, no longer used, VSD jobs.
* Mark constants as `Final`.

Re: mark constants as `Final`. If you originally did not use [upper case names for you constants](https://www.python.org/dev/peps/pep-0008/#constants), "your" constants will not have been touched. There's only so much regex magic I was willing to let loose on our code base.
20210624_124625_screenshot


jjmurre (approved):
LGTM, again a nice set of cleanups!

cvriel (approved):
Wow, mooie opschoning :-)

PR #292, opened 2021-06-28
OGR2OGR promote_to_multi fix

cvriel:
The promote_to_multi option was not optional yet. Turn it into an boolean value where de default is set to False (not promoting geometry objects to multigeometry objects).

Currently it was leading to errors in Airflow because it was defaulted to always promote. This lead to geometry check issues.


jjmurre (approved):
👍🏼 Lijkt me prima zo!

PR #294, opened 2021-06-28
Hotfix: remove debugging code from csma DAG.

gkoller:
After review feedback I had to change a couple of things that I tested
with this debugging code in place. Though I had forgotten to remove it
in the final push.

### DIFF
--- a/src/dags/cmsa.py
+++ b/src/dags/cmsa.py
@@ -1,7 +1,6 @@
 from pathlib import Path
 from typing import Final
 
-import pendulum
 from airflow import DAG
 from airflow.models import Variable
 from airflow.operators.dummy import DummyOperator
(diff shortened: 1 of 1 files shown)


src/dags/cmsa.py
```diff
@@ -226,8 +224,3 @@
     Example geosearch:
     https://api.data.amsterdam.nl/geosearch?datasets=cmsa/locatie&x=106434&y=488995&radius=10
 """
-if __name__ == "__main__":
-    from airflow.utils.state import State
-
-    dag.clear(dag_run_state=State.NONE)
-    dag.run()
```
ngaranko, line 233: nice.

gkoller: Together with the `DebugExecutor` this allows you to run the module containing the DAG as a regular Python program. Full power of PyCharm's debugger at your disposal. No need to run a separate `airflow scheduler` process.

PR #300, opened 2021-07-06
AB#8751 hr updates

jjmurre:
Two changes caused by adding/changing shortnames to HR dataset.

### DIFF
--- a/src/plugins/sqlalchemy_create_object_operator.py
+++ b/src/plugins/sqlalchemy_create_object_operator.py
@@ -9,7 +9,7 @@ from environs import Env
 from more_ds.network.url import URL
 from schematools.cli import _get_engine
 from schematools.importer.base import BaseImporter
-from schematools.utils import schema_def_from_url
+from schematools.utils import schema_def_from_url, to_snake_case
 from xcom_attr_assigner_mixin import XComAttrAssignerMixin
 
 env = Env()
@@ -36,7 +36,7 @@ class SqlAlchemyCreateObjectOperator(BaseOperator, XComAttrAssignerMixin):
         "data_table_name",
     ]
 
-    @apply_defaults
+    @apply_defaults  # type: ignore [misc]
     def __init__(
         self,
         data_schema_name: str,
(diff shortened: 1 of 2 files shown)


src/plugins/sqlalchemy_create_object_operator.py
```diff
@@ -135,7 +135,7 @@ def execute(self, context: Dict[str, Any]) -> None:
             self.log.info("Considering table '%s'.", table.name)
             cur_table = f"{self.data_schema_name}_{table.name}"
 
-            if re.fullmatch(self.data_table_name, cur_table):
+            if re.fullmatch(self.data_table_name, to_snake_case(cur_table)):
```
gkoller, line 138: Solves the problem :+1: 

But it is yet another `to_snake_case` application. We still need to come up with a good solution to attack the proliferation of the `to_snake_case` usages in our code base and at the same time the occasional accidental omission causing bugs like this PR is solving.

Ignoring the brainstorming session we had on Slack the other day about using Pydantic for a complete rewrite of the Amsterdam Schema representation in Python and whatever a solution may come out of that. Maybe we can do something like the following in the mean-time:

Either: 
Define a `__getattr__` on `SchemaType` that translates data and property attribute references that start with `sc_` to return the snake cased variant of the attribute:

```python
    def __getattr__(self, name: str) -> Any:
        if name.startswith("sc_"):
            attr = getattr(self, name[len("sc_"):])
            if isinstance(attr, str):
                # only works for data attributes & properties, not regular methods
                return to_snake_case(attr)
        raise AttributeError(f"{self.__class__.__name__!r} object has no attribute {name!r}")
```
This allows you to do `table.sc_name`. But his is probably just as bad as using `to_snake_case` directly.

or better:
Let the `SchemaType` derived classes remain the JSON representation (with CamelCase names). Have another set of data structures that accept a SchemaType derived instance, but wit all attributes references returning snake cased values.

or even better:
JSON schemas remain in CamelCase but as soon as we read in the JSON schema into a Python `SchemaType` derived datastructure convert everything once to snake case. After all we are only dealing with snake case from that moment on. Hence it is best to do the conversion immediately. Meaning that `table.name` would have returned a snake cased name.

gkoller: BTW I'm not advocating this to be solved in this PR. It is just that this PR triggered my `to_snake_case` trigger. ;-). PR is good to go.

gkoller (approved):

src/plugins/sqlalchemy_create_object_operator.py
jjmurre: Totally agree. I like the "even better" approach most. In that case we have a clear "interface" entering or exiting the snake_case world.

PR #302, opened 2021-07-07
bouwstroompunten fix

cvriel:
The bouwstroompunten uses a HttpFetchOperator to call an API to retrieve source data.For using the API a API token is needed. This token was added to the header. By adding a get_token() within the header definition.

However during 'compile' (when importing the dags) the get_token() was evaluated. Resulting an a error message because the parameters used are not yet loaded.

To prevent from evaluting at compile time (instead of only at run time) the output of get_token() is now passed as a xcom variable. Based on the same logic as BBGA for consistency.

### DIFF
--- a/src/dags/bouwstroompunten.py
+++ b/src/dags/bouwstroompunten.py
@@ -7,6 +7,7 @@ import dsnparse
 import requests
 from airflow import DAG
 from airflow.models import Variable
+from airflow.operators.python import PythonOperator
 from airflow.providers.postgres.operators.postgres import PostgresOperator
 from common import (
     DATAPUNT_ENVIRONMENT,
@@ -45,7 +46,7 @@ geo_checks = []
 db_conn: object = DatabaseEngine()
 
 
-def get_token() -> Union[str, Any]:
+def _get_token() -> Union[str, Any]:
     """Getting the access token for calling the data endpoint.
 
     Returns:
@@ -55,9 +56,6 @@ def get_token() -> Union[str, Any]:
     `AIRFLOW_CONN_BOUWSTROOMPUNTEN_BASE_URL` the scheme en host must be extracted
     and merged into a base endpoint.
     """
-    if not user or not password:
-        raise ValueError("no credentials available")
-
     scheme = dsnparse.parse(base_url).scheme
     host = dsnparse.parse(base_url).hostloc
     base_endpoint = "".join([scheme, "://", host])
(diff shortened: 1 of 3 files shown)


src/plugins/http_fetch_operator.py
```diff
@@ -68,15 +74,27 @@ def __init__(
...
         self.output_type = output_type  # default is raw, else specify text i.e.
         self.xcom_tmp_dir_task_ids = xcom_tmp_dir_task_ids
-        self.xcom_tmp_dir_key = xcom_tmp_dir_key
+        self.xcom_key = xcom_key
+        self.xcom_token_task_ids = xcom_token_task_ids
         self.verify = verify
 
         super().__init__(*args, **kwargs)
 
     def execute(self, context: Dict) -> None:  # noqa: C901
```
jjmurre, line 83: `Context` heeft zelfs netjes een eigen type in Airflow:
    from airflow.models.taskinstance import Context

src/dags/bouwstroompunten.py
```diff
@@ -45,7 +46,7 @@
 db_conn: object = DatabaseEngine()
 
 
-def get_token() -> Union[str, Any]:
+def _get_token() -> Union[str, Any]:
```
gkoller, line 49:
```python
from airflow.decorators import task
...
@task  # type: ignore[misc]
def get_token() -> str:
   ...
```

src/dags/bouwstroompunten.py
```diff
@@ -102,15 +100,22 @@ def get_token() -> Union[str, Any]:
     # 2. Create temp directory to store files
     mkdir = mk_dir(tmp_dir, clean_if_exists=False)
 
+    # GET TOKEN
+    get_token = PythonOperator(
```
gkoller, line 104: Now that `get_token` (important: without the `_` prefix) is a `@task` we don't need the `PythonOperator` anymore. This is the Airflow Flow 2.0 API. Much more convenient than the old one. Whatever a `@task` returns is stored in XCOM under `xcom_key` with the function name as the value: `"get_token"` in this case. That's the reason I stripped the `_` prefix. After all `HTTPFetchOperator` expect the token to be retrieveable under the `xcom_key` `"get_token"`
```python
    get_jwt_token = get_token()
```

src/dags/bouwstroompunten.py
```diff
@@ -191,6 +196,7 @@ def get_token() -> Union[str, Any]:
     (
         slack_at_start
         >> mkdir
+        >> get_token
```
gkoller, line 199: `get_jwt_token`

jjmurre (approved):
LGTM. Werkt mooi met die nieuwe xcom feature van Airflow 2.

PR #305, opened 2021-07-12
Guido/ab8702 convert basiskaart to schema based table creation

gkoller:
This PR creates basiskaart tables based on the Amsterdam Schema. To do so, this has been turned into a single static DAG instead of a dynamically generated DAG per basiskaart table as it used to be.

As always some code cleanups, and as always, best reviewed per commit.

  * [AB#8701](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8701) Simplify copy_data_in_batch and add TODO.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Fix inconsistent formatting of SQL SELECT stmts.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Remove CREATE TABLE SQL statements.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Schema based table creation for basiskaart.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Do specify column names for DB inserts.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Rewrite basiskaart to be one DAG with parallel tasks.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Clean up LogMessageOperator.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Remove always true condition 1=1 in SQL stmts.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Support BSD fs hierarchy. Subprocess error handling.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Use logger defined on class.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) capture output in order to be able to display it.
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Call format on `SQL` not on `str`
  * [AB#8702](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/8702) Capitalize constants.


src/dags/basiskaart.py
```diff
@@ -36,12 +36,12 @@
...
 
-dag_id = "basiskaart"
-owner = "dataservices"
-variables = Variable.get(dag_id, deserialize_json=True)
-tables_to_create = variables["tables_to_create"]
-source_connection = "AIRFLOW_CONN_POSTGRES_BASISKAART"
-import_step = 10000
+DAG_ID: Final = "basiskaart"
+OWNER: Final = "dataservices"
+VARIABLES: Final = Variable.get(DAG_ID, deserialize_json=True)
```
jjmurre, line 45: The `Variable.get` is doing a db-query. So, `VARIABLES` is not a real constant that is fully defined in code. But, well, what is the definition of a constant?

src/plugins/provenance_rename_operator.py
```diff
@@ -186,14 +185,10 @@ def execute(self, context: Context) -> None:
...
                                     RENAME TO {new_index_name}
-                            """.format(
-                                    old_index_name=sql.Identifier(
-                                        self.pg_schema, index_name
-                                    ).as_string(connection),
-                                    new_index_name=sql.Identifier(new_index_name).as_string(
-                                        connection
-                                    ),
-                                )
+                            """
```
jjmurre, line 188: 👍🏼 Nice, this is the real fix :-)

src/plugins/ogr2ogr_operator.py
```diff
@@ -85,7 +85,8 @@ def execute(self, context: Context) -> None:
         input_file.parents[0].mkdir(parents=True, exist_ok=True)
 
         # setup the cmd to execute
-        ogr2ogr_cmd: List[str] = ["ogr2ogr", "-f", self.mode]
+        program = "ogr2ogr"
```
jjmurre, line 85: What is the value of defining a variable, when it is only used once?

src/plugins/ogr2ogr_operator.py
```diff
@@ -137,7 +134,7 @@ def execute(self, context: Context) -> None:
             result = subprocess.run(  # noqa: S603
                 ogr2ogr_cmd, capture_output=True, check=True, text=True
             )
-            logger.debug(result.stdout)
+            self.log.debug(result.stdout)
```
jjmurre, line 137: Nice, consistency!

src/plugins/pgcomparator_cdc_operator.py
```diff
@@ -85,14 +88,17 @@ def execute(self, context: Dict[str, Any]) -> None:
...
+        )
         self.log.info(
-            "Start change data capture (with pg_comparator) from: '%s' to: '%s" ".",
+            "Start change data capture (with pg_comparator) from: '%s' to: '%s.",
             source_url,
             target_url,
         )
         arguments = [
-            "/usr/bin/pg_comparator",
+            program_path.as_posix(),
```
jjmurre, line 101: Looking at the git history, this was oringinally a command without a path. There should have been a reason to change it.

src/plugins/log_message_operator.py
```diff
@@ -1,33 +1,34 @@
...
         self.message = message
         self.channel = channel
-        super().__init__(*args, **kwargs)
 
-    def execute(self, context):
+        # Only pass along arguments that the BaseOperator super class defines.
+        # Reason being that by default BaseOperator does not accept arguments it didn't
+        # define. Eg arguments that are specific to the class we are proxying, the
+        # SlackWebhookOperator.
+        sig = inspect.signature(BaseOperator.__init__)
```
jjmurre, line 25: Nice approach!

src/dags/basiskaart.py
lbam: A constant is a thing that always has the same value. That fits the bill :)

src/dags/basiskaart.py
```diff
@@ -155,84 +147,84 @@ def create_basiskaart_dag(is_first: bool, table_name: str, select_statement: str
...
-        # 8. Trigger next DAG to run (estafette)
-        trigger_next_dag = TriggerDynamicDagRunOperator(
-            task_id="trigger_next_dag",
-            dag_id_prefix=f"{DAG_ID}_",
-            trigger_rule="all_done",
-        )
-
-        # 9. Grant database permissions
-        grant_db_permissions = PostgresPermissionsOperator(task_id="grants", dag_name=DAG_ID)
+    grant_db_permissions = PostgresPermissionsOperator(task_id="grants", dag_name=DAG_ID)
```
jjmurre, line 233: The `TriggerDynamicDagRunOperator` has been dropped now. However, that was meant to serialize the DAGs, to avoid some "thundering herd" of DAGs. According to the main comment of this commit, everthing has been parallelized now. So, be carefull to check performance when this code goes to production.

src/dags/sql/basiskaart.py
```diff
@@ -14,89 +14,6 @@
 # necessary materialized views (see note 1) for the vector based tile generator.
 # i.e. CREATE MATERIALIZED VIEW <name_view> AS SELECT * FROM <table_name>
 #
```
jjmurre, line 17: Nice to get rid of these!

jjmurre (approved):
LGTM. One issue to be careful with is the serialized vs parallel execution of the DAGs.

src/plugins/pgcomparator_cdc_operator.py
ngaranko: @jjmurre I guess FreeBSD came into play here with `/usr/local/bin` :)

src/plugins/ogr2ogr_operator.py
gkoller: It is used twice. Second usage in `logger.exception("Failed to run %r.", program)`

src/plugins/pgcomparator_cdc_operator.py
gkoller: @jjmurre I'm stupid. I introduced the path component in 84a507d7e7734eaf397e6037d418dab2fda6f169 and didn't document the need for the change. Will revert it back to being pathless as that should work equally well.

@ngaranko Yep :-)

src/dags/basiskaart.py
gkoller: Ah, I see. I was initially in doubt how to deal with `TriggerDynamicDagRunOperator`. Thing is that `SqlAlchemyCreateObjectOperator` creates the tables for all dynamic DAGs in one go. Felt a bit weird to have that executed in each dynamic DAG. Executing it only once, namely in the first dynamic DAG being run, felt awkward too, as I feel that each DAG should list all the tasks in needs to achieve its goal. As such creating a dependency between DAG didn't feel right

Will keep an eye on the performance. Especially now that I have seen this setting in `airflow.cfg`:
```
# The amount of parallelism as a setting to the executor. This defines
# the max number of task instances that should run simultaneously
# on this airflow installation
parallelism = 32
```

src/plugins/pgcomparator_cdc_operator.py
gkoller: Turns out I am not stupid. Or at least, less stupid than I'd thought. In commit 84a507d7e7734eaf397e6037d418dab2fda6f169 I ditched `shell=True` in the `subprocess.run` call. Hence the need for the full path to the executable. Can't rely on a shell to inspect the PATH environment variable as we are not using the shell anymore.

lbam: `subprocess.run` checks the `PATH` even when `shell=False`. Try `subprocess.run(["ls"], shell=False)`.

src/dags/basiskaart.py
gkoller: You got me sufficiently worried that I have tried to address the problem by creating a subdag per table. This now (still needs to be commited) looks like:
20210713_155316_screenshot

gkoller: `SubDag`'s are a pain. So that solution didn't work. I have now serialized all tasks and used `TaskGroup`s to visually partition the sub units of work in the GUI. This is how it looks;
20210713_171835_screenshot

It should prevent the thundering herd problem (overwhelming the DB).

src/plugins/pgcomparator_cdc_operator.py
gkoller: Ah, didn't know that. Will fix in next PR. Thank you @lbam . Learned something new.

PR #307, opened 2021-07-14
AB#4378 drop target if unequal

jjmurre:
See commit message for explanation.

### DIFF
--- a/src/plugins/postgres_table_copy_operator.py
+++ b/src/plugins/postgres_table_copy_operator.py
@@ -8,6 +8,7 @@ from airflow.exceptions import AirflowFailException
 from airflow.models import XCOM_RETURN_KEY, BaseOperator
 from airflow.providers.postgres.hooks.postgres import PostgresHook
 from airflow.utils.decorators import apply_defaults
+from more_itertools import first
 from psycopg2 import sql
 from xcom_attr_assigner_mixin import XComAttrAssignerMixin
 
@@ -44,12 +45,13 @@ class PostgresTableCopyOperator(BaseOperator, XComAttrAssignerMixin):
        configurable.
     """
 
-    @apply_defaults
+    @apply_defaults  # type: ignore [misc]
     def __init__(
         self,
         source_table_name: Optional[str] = None,
         target_table_name: Optional[str] = None,
         truncate_target: bool = True,
+        drop_target_if_unequal: bool = False,
         copy_data: bool = True,
         drop_source: bool = True,
         task_id: str = "copy_table",
@@ -68,6 +70,8 @@ class PostgresTableCopyOperator(BaseOperator, XComAttrAssignerMixin):
             source_table_name: Name of the temporary table that needs to be copied.
             target_table_name: Name of the table that needs to be created and data copied to.
             truncate_target: Whether the target table should be truncated before being copied to.
+            drop_target_if_unequal: Whether the target table should be dropped if it is not
+                equal to the source table wrt. column types, names and positions
             copy_data: Whether to copied data. If set to False only the table definition is copied.
             drop_source: Whether to drop the source table after the copy process.
             task_id: Task ID
@@ -86,6 +90,7 @@ class PostgresTableCopyOperator(BaseOperator, XComAttrAssignerMixin):
         self.source_table_name = source_table_name
         self.target_table_name = target_table_name
         self.truncate_target = truncate_target
+        self.drop_target_if_unequal = drop_target_if_unequal
         self.copy_data = copy_data
         self.drop_source = drop_source
         self.postgres_conn_id = postgres_conn_id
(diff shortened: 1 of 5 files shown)


src/plugins/postgres_table_copy_operator.py
```diff
@@ -112,6 +116,50 @@ def __init__(
...
+               WHERE table_name = %(source_table_name)s)
+            """
+                ),
+                (
+                    {
+                        "source_table_name": table_mapping.source,
+                        "target_table_name": table_mapping.target,
+                    }
+                ),
+            )
```
lbam, line 154: Why the extra parentheses?

src/plugins/postgres_table_copy_operator.py
```diff
@@ -112,6 +116,50 @@ def __init__(
...
+                      ordinal_position,
+                      column_name
+               FROM information_schema.columns
+               WHERE table_name = %(target_table_name)s
+               EXCEPT SELECT data_type,
+                             ordinal_position,
+                             column_name
+               FROM information_schema.columns
+               WHERE table_name = %(source_table_name)s)
+            """
```
lbam, line 146: You can do this in less code with a full outer join:

```sql
WITH src AS (
    SELECT data_type, ordinal_position, column_name FROM information_schema.columns
    WHERE table_name = %(source_table_name)s
), tgt AS (
    SELECT data_type, ordinal_position, column_name FROM information_schema.columns
    WHERE table_name = %(source_table_name)s
)
SELECT (COUNT(*) = COUNT(src)) FROM src NATURAL FULL OUTER JOIN tgt
```

src/plugins/postgres_table_copy_operator.py
jjmurre: Ah, that was a left-over from editing stuff, removed! Tx.

src/plugins/postgres_table_copy_operator.py
jjmurre: That's a nice improvement, I'll replace the query with this one.

lbam: Please test it first, I didn't test this final version with the string interpolation.

jjmurre: Yes, I tested it. There turned out the be one flow. If the target table was smaller, it did not work, so I changed the condition in the select to:

                        SELECT COUNT(*) = COUNT(src) AND count(*) = COUNT(tgt)

jjmurre: And during testing, I discovered a bug. There was no check on the existence of the target table, so, I added an extra check.

lbam: Oh, good catch.

src/plugins/postgres_table_copy_operator.py
```diff
@@ -119,40 +120,44 @@ def __init__(
     def _drop_tables_if_unequal(self, cursor: Any, all_table_copies: List[TableMapping]) -> None:
         """Drop the target tables that are unequal to their source table."""
         for table_mapping in all_table_copies:
+            # First make sure the target table exists
+            cursor.execute(
+                sql.SQL("SELECT to_regclass('{target_table_name}')").format(
+                    target_table_name=sql.Identifier(table_mapping.target)
+                )
+            )
```
lbam, line 128: Are you sure this is necessary? When I run the outer join query with a non-existing table name, I get a false result (assuming the other table exists).

lbam (approved):

src/plugins/postgres_table_copy_operator.py
jjmurre: The problem is, false result is interpreted as tables are not equal. So next step is to remove the target table. However, that does not exist, leading to crash in postgres.

lbam: Oh, right. Ok!

src/plugins/postgres_table_copy_operator.py
```diff
@@ -112,6 +117,54 @@ def __init__(
                 "even though source table will be dropped."
             )
 
+    def _drop_tables_if_unequal(self, cursor: Any, all_table_copies: List[TableMapping]) -> None:
+        """Drop the target tables that are unequal to their source table."""
+        for table_mapping in all_table_copies:
+            # First make sure the target table exists
+            cursor.execute(
+                sql.SQL("SELECT to_regclass('{target_table_name}')").format(
```
gkoller, line 125: I'm very glad to see `psycopg2.sql` used in more and more code in favor of `str` formatting.

Here however a regular DBAPI parameter passing would have sufficed. Eg:

```python
cursor.execute(
    "SELECT to_regclass(%(target_table_name)s)",
    {target_table_name=table_mapping.target}
)
```

Reason being that `target_table_name` is not really an identifier from the SQL grammar point of view. Yes, it is a table name and table names are identifiers, but here it is just a string argument to a PostgreSQL function. Simply using `%(target_table_name)s` will make sure it is quoted as a string argument (hence you can leave the quotes out).

And as we're just using a regular SQL query with plain DBAPI parameters, you can do away with `sql.SQL` and `sql.Identifier`

src/plugins/postgres_table_copy_operator.py
```diff
@@ -112,6 +117,54 @@ def __init__(
...
+                )
+            )
+
+            if first(cursor.fetchone()) is None:
+                self.log.info(
+                    "Target table %s not found, skipping equality check.", table_mapping.target
+                )
+                continue
+            cursor.execute(
+                sql.SQL(
```
gkoller, line 136: No need for `sql.SQL` here. You are not doing anything fance with `sql.Identifier` or `sql.Placeholder` or whatever. This is just a regular SQL query with DBAPI parameters (the `%(...)s ones).

BTW I like this query. Nice solution.

src/plugins/postgres_table_copy_operator.py
```diff
@@ -112,6 +117,54 @@ def __init__(
...
+                {
+                    "source_table_name": table_mapping.source,
+                    "target_table_name": table_mapping.target,
+                },
+            )
+            if not first(cursor.fetchone()):
+                self.log.info("Dropping table %s", table_mapping.target)
+                cursor.execute(
+                    sql.SQL("DROP TABLE {table_name}").format(
+                        table_name=sql.Identifier(table_mapping.target)
```
gkoller, line 164: Yes! :+1:  Here the table name is indeed an identifier from the SQL grammar point of view. Hence you need to use `postgresql.sql` helpers.

gkoller (approved):
Very nice addition. Adding this here allows us to remove many `rm_tmp_table` tasks in our DAGs :+1:

PR #308, opened 2021-07-14
sport fix

cvriel:
In the SQL update statement the WHERE clause was omitted from the logic. Resulting in a invalid statement. 

- Added the WHERE in the update statement


gkoller (approved):
LGTM.

I'm sorry for having introduced this bug in the first place.

PR #319, opened 2021-07-26
Change cmsa view from v8 -> v10

jjmurre:


cvriel (approved):
Yeay that is all I quess, will look at openstack for the select grants

PR #332, opened 2021-08-02
[WIBON data] first pipeline

cvriel:
Setup of the WIBON datapipeline. Complementairy to exisiting dataset leidingeninfractuur. The latter consisting of tables which are filled by the datateam Ruimte. The extra new tables WIBON are added by Airflow.


ngaranko (approved):
LGTM

PR #340, opened 2021-08-20
Remove use of pg-comparator

jjmurre:
See info info in the commit messages.


lbam:
Are these all the `pg_comparator` uses in the project? If so, we can remove it from `README.md` and the various `Dockerfiles`.

jjmurre:
> Are these all the `pg_comparator` uses in the project? If so, we can remove it from `README.md` and the various `Dockerfiles`.

Yes. You are right. I waited with that, because Chris/Bas are moving Airflow to Azure and constructed a new Dockerfile. I have to ask them if that will fully replace the current one. And, I have to talk to @gkoller about the BBGA DAG, maybe that DAG can be simplified a bit, now that PG comparator has been removed.

cvriel (approved):
Met ook nog verbeteringen t.a.v. mypy, flake8 en black :+1:

gkoller:
>  And, I have to talk to @gkoller about the BBGA DAG, maybe that DAG can be simplified a bit, now that PG comparator has been removed.

Yes, it can be simplified quite a bit. A lot of its processing is generating a unique key for PG Comparator to be working with.

However, I am wondering whether we should be removing PG Comparator at all. It has a clear benefit over truncate/insert. Namely in that it can keep `id`'s the same in case of mere updates. With truncate/insert `id`'s might, no *will* change, if new rows are inserted in the source dataset anywhere but at the end.

Changing `id`'s that DSO-API users depends on is bad. Not to mention it violates the whole concept of `id`'s being IDs.

gkoller:
> However, I am wondering whether we should be removing PG Comparator at all.

I do realize PG Comparator is not supported by Azure managed Postgres, however with Terraform notthing prevents us from spinning up our own database. I means that's one of the reasons for using Terraform, right?

gkoller (approved):

gkoller:
> > However, I am wondering whether we should be removing PG Comparator at all.
> 
> I do realize PG Comparator is not supported by Azure managed Postgres, however with Terraform nothing prevents us from spinning up our own database. I mean that's one of the reasons for using Terraform, right?

We've discussed this briefly during our meeting this morning. I do understand the need for removing the operator at this time.

jjmurre:
> > Are these all the `pg_comparator` uses in the project? If so, we can remove it from `README.md` and the various `Dockerfiles`.
> 
> Yes. You are right. I waited with that, because Chris/Bas are moving Airflow to Azure and constructed a new Dockerfile. I have to ask them if that will fully replace the current one. And, I have to talk to @gkoller about the BBGA DAG, maybe that DAG can be simplified a bit, now that PG comparator has been removed.

Every pg_comparator related thing has now been removed.

PR #341, opened 2021-08-24
Guido/refactor bbga fix perf

gkoller:
Refactor BBGA to:
- use table rename instead of table copy
- use SQLite for transforming CSV files instead of slooowwww pure Python code

*IMPORTANT*
Before merging this PR we should drop the `cdc_id` columns on all `bbga_` tables in `acc` and `prod`.


jjmurre (approved):
LGTM. Clever use of SQLite with a massive performance gain!

jjmurre:
Is the removal of `cdc_id` from the database really needed? Because we now use the `TableRenameOperator`, so we "swap in" a whole new fresh table?

src/dags/sql/bbga_sqlite_transform.py
```diff
@@ -0,0 +1,110 @@
...
+ we need to create our own. Quite sad. However this whole SQLite3
+ exercise is because of the requirement to derive a primary key.
+ SQLite3 can not only easily import and export CSV files, it executes
+ these transformations so much faster than pure Python code.
+ */
+ALTER TABLE bbga_kerncijfers
+    ADD COLUMN id VARCHAR GENERATED ALWAYS AS (indicator_definitie_id || '|' || jaar || '|' || gebiedcode_15) VIRTUAL;
+
+ALTER TABLE bbga_statistieken
+    ADD COLUMN id VARCHAR GENERATED ALWAYS AS (indicator_definitie_id || '|' || jaar) VIRTUAL;
```
lbam, line 92: Can/should we put NOT NULL constraints on these?

lbam:
Brilliant, Too bad SQLite doesn't have crypto hashes built-in, but the new identifiers should be fine too.

PR #344, opened 2021-09-01
README update

mbr564:
2 minor changes in the README file:

- Python 3.8 is required for some packages, so the python 3.5 - 3.7 requirement had to be changed
- Airflow command to update variables is changed in newer versions of Airflow


jjmurre (approved):
LGTM

PR #346, opened 2021-09-02
Add support for populating new table bbga_cijfers.

gkoller:
`bbga_cijfers` is a convenience table in order to make the dashboard
https://gebiedinbeeld.amsterdam.nl/ performant. Haven't seen the
frontend code, hence can't judge the architecture, but they want to
retrieve both `kerncijfers` and `statistieken` data in one API call. And
this was the easiest solution.

See also https://github.com/Amsterdam/amsterdam-schema/commit/5ffa2bbac1317e391a9c943135c84aaf9de965d1

### DIFF
--- a/src/dags/bbga.py
+++ b/src/dags/bbga.py
@@ -1,6 +1,6 @@
 import shutil
 from pathlib import Path
-from typing import Dict, Final
+from typing import Dict, Final, Tuple
 
 from airflow import DAG
 from airflow.models import Variable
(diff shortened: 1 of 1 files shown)


src/dags/bbga.py
```diff
@@ -35,11 +36,14 @@
 data_endpoints: Dict[FileStem, UrlPath] = VARS["data_endpoints"]
 table_mappings: Dict[FileStem, str] = VARS["table_mappings"]
 
+# We have one extra table that does not map to a downloadable dataset
+# Hence we add it explicitly here.
+TABLES: Tuple[str, ...] = tuple(chain(table_mappings.values(), ("bbga_cijfers",)))
```
lbam, line 41: Surely this is not performance-critical. Can we use `+` concatenation for readability?

```suggestion
TABLES: Tuple[str, ...] = tuple(table_mappings.values()) + ("bbga_cijfers",)
```

jjmurre (approved):
LGTM, exactly as we already discussed!

src/dags/bbga.py
gkoller: Wasn't after performance. This is just how list/tuple concatenation based code springs into existence in my brain. 😉

Will change and amend existing commit once I'm back at my laptop

PR #348, opened 2021-09-07
swiftoperator explicit connection to VSD objectstore

cvriel:
In some DAG''s the swiftoperator is used to download data from the VSD objectstore. The parameter `swift_conn_id` is in many cases omitted. Because with the parameter it defaults to the VSD objectstore. However this default swift (objectstore) connection is not using the baseoperator.get_connection(). This has a drawback on using KeyVault on the Airflow instance on AKS. Basicly KeyVault is not consulted. When ommitted it will use the swift API that will look for OS_* env vars. The Airflow on AKS based on a HELM installation is expecting the Airflow naming conventions for env vars. Hence, that would be AIRFLOW_CONN_*

To solve this, every DAG should always explicitly add swift_id as the `swift_conn_id`. Then the Airflow native baseoperator.get_connection() will be used.

### DIFF
--- a/src/dags/asbest.py
+++ b/src/dags/asbest.py
@@ -51,6 +51,7 @@ with DAG(
 
     fetch_zip = SwiftOperator(
         task_id="fetch_zip",
+        swift_conn_id="SWIFT_DEFAULT",
         container=dag_id,
         object_id=zip_file,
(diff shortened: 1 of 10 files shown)


jjmurre (approved):
LGTM. According to the zen of python "Explicit is better than implicit" :-)

PR #353, opened 2021-09-14
Modified branch selector in stroomstoringen DAG

mbr564:
Previously, the path in the DAG was chosen based on the success/failure of the 'check_table' PostgresOperator. This worked, but in case of failure of the check_table operator, a failure message was shown in the slack group. This was confusing. I replaced this PostgresOperator with a BranchPythonOperator that checks if the retrieved .geojson file contains data on power failures. Based on the number of power failures it defines the wanted path:
1.  1 or more power failures -> modify and store data
2.  no power failures -> create dummy data

### DIFF
--- a/src/dags/stroomstoringen.py
+++ b/src/dags/stroomstoringen.py
@@ -59,7 +59,7 @@ with DAG(
     # 2. Create temp directory to store files
     mkdir = mk_dir(TMP_DIR, clean_if_exists=False)
 
-    # 3. download the data into temp directoys
+    # 3. download the data into temp directories
     download_data = HttpFetchOperator(
         task_id="download",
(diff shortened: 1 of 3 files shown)


src/dags/stroomstoringen.py
```diff
@@ -63,7 +59,7 @@
     # 2. Create temp directory to store files
     mkdir = mk_dir(TMP_DIR, clean_if_exists=False)
 
-    # 3. download the data into temp directorys
+    # 3. download the data into temp directoys
```
jjmurre, line 62: Small typo: directories

jjmurre (approved):
LGTM, nice solution using branching.

PR #356, opened 2021-09-21
bedrijveninvesteringszones fix

markdb-Adam:
The error occurs when pandas is used in python3.9+, cause the the code:
xml.etree.ElementTree.Element.getiterator() has been removed.

A workaround is to utilize the openpyxl engine to read the excel files. So, adding openpyxl to the requirements and changing pd.read_excel('file.xlsx') to pd.read_excel('file.xlsx', engine='openpyxl') solved the issue.


jjmurre (approved):
Looks good! Can you check if cmsa error has also been solved by this fix.

PR #357, opened 2021-09-27
Removed API restriction to show all data

cvriel:
The stroompunten source API was not fully showing all data with a single API call. By using the _limit=-1 variable the data can be fully procesessed. Due to this output limit not all stroompunten where visible at data.amsterdam.nl.


markdb-Adam (commented):
Looking good to me!

markdb-Adam (approved):

PR #372, opened 2021-11-02
BBGA: Fix: set PK to `variabele` instead of `sort`.

gkoller:
`Variabele` is the correct primary key. That is also how it is defined
in the Amsterdam Schema. Not sure why I originally set it to `sort` for
the duration of the import of the CSV.


jjmurre (approved):
LGTM

PR #376, opened 2021-11-09
due to source change, datatype had to changed from integer to numeric

cvriel:
The range of the attribute zaaknummer has changed so it out ranges an integer datatype. It was therefore modified to bigint.


src/dags/horeca_exploitatievergunning.py
```diff
@@ -87,7 +87,7 @@ def load_from_dwh(table_name: str, source_srid: int) -> None:
             func=wkt_loads_wrapped, source_srid=source_srid
         )
         dtype = {
-            "zaaknummer": Integer(),
+            "zaaknummer": Numeric(),
```
jjmurre, line 90: I guess this is really an integer, right, not a floating point number. Wouldn't it be better to use a bigint: https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.BigInteger

PR #378, opened 2021-11-11
beheerkaart not using correct database objects names

cvriel:
Since the change in Amsterdam schema definition of beheerkaart, the dataset uses a camelCase as in `beheerkaartBasis`. The DAG that utilizes operators that for instance create a database object needs a snake cased equivalent. Also due to creating of objects in a database schema other then `public` as in `pte` the creating of tables in the explicit correct database schema was not in place leading to errors.


jjmurre (approved):
There it is again, the infamous snake/camel controversy :-)

PR #383, opened 2021-11-30
bugfix in check on geometry

cvriel:
Classic bug that prevents a check on geometry data. The code was reusing a list that holds the checks for each dataset, in stead of creating a list per dataset. Consequence was that the last check of a particular dataset was executed for all datasets. Masking issues on other datasets.


jjmurre (approved):
LGTM. Yeah, mutabilty is a bitch.

PR #395, opened 2022-01-19
Fix a know bug in Airflow using a patch with the upstream fix

jjmurre:
Upgrading (from 2.1.4 -> 2.2.3) Airflow is not possible atm.
according to team dataverwerking. They are taking this up
as part of their roadmap.

So, for now this patch should suffice and should be removed
after the upgrade of Airflow.


src/patches/views.py.patch
```diff
@@ -0,0 +1,22 @@
+--- /usr/local/lib/python3.9/site-packages/airflow/www/views.py 2022-01-19 10:47:20.035044670 +0000
++++ /usr/local/lib/python3.9/site-packages/airflow/www/views.py 2022-01-19 10:50:02.584152438 +0000
+@@ -849,11 +849,18 @@
+         # Filter to only ask for accessible and selected dags
+         query = query.filter(DagRun.dag_id.in_(filter_dag_ids))  # pylint: enable=no-member
+
++        # Pathed version from upstream Airflow
```
lbam, line 7:
```suggestion
+        # Patched version from upstream Airflow
```

lbam (approved):

PR #411, opened 2022-02-21
[FIX] PostgresPermissionsOperator was not latest version

cvriel:
Due to the new situation of having level nested datasets, like in dataset/table and dataset/sub-dataset/table, multiple levels for finding tables had to defined in the tables lookup for database grants. However this new change was not part of the last commit. So theresfor it was added just now.


jjmurre (approved):
LGTM

PR #416, opened 2022-02-24
Fix table-init and -copy operator to handle nested tables

jjmurre:
In the PostgresTableInitOperator and the PostgresTableCopyOperator
the related tables are found based on the name of the base table.

However, when the base tablename has as `_new` postfix, this
approach does not work for the nested tables.

This has now been solved by passing the names of these nested tables
explicitly to these two operators.


cvriel (approved):
Nog best een flinke refactor! Goed dat we nu expliciet de DB nested table namen kunnen toevoegen.

PR #418, opened 2022-03-02
Fix GOB scheduling

jjmurre:
The dynamic `start_date` that was being used (two days in the past)
is not compatible with less frequently running DAGs.

So, a fixed `start_date` is being used (which is also recommended
by the Airflow docs).

Furthermore, the scheduling setup was not compatible with new
requests by BenK. Some datasets needed to be in several estafette runs.


cvriel (approved):
LGTM

PR #429, opened 2022-04-04
[UPDATE] slack post message

cvriel:
Currently posting Slack messages is bound to one channel (#airflow). With regards to onboarding multiple datateams, willing to post their messages on their own channel, the way Slack messages was currently implemented needed to be updated. Also moving the code from `common` to `plugins` was also included to keep all operators under the same folder (plugins) for consistency.


src/dags/rioolnetwerk.py
```diff
@@ -4,13 +4,10 @@
 from airflow import DAG
 from airflow.providers.postgres.operators.postgres import PostgresOperator
 from common import (
-    OTAP_ENVIRONMENT,
     DATASTORE_TYPE,
-    SLACK_ICON_START,
     MessageOperator,
```
RonvanBarneveld, line 8: No SHARED_DIR needed?

src/dags/vergunningen.py
```diff
@@ -8,11 +8,9 @@
 from common import (
     OTAP_ENVIRONMENT,
```
RonvanBarneveld, line 9: Why is the OTAP_ENVIRONMENT needed here? In the most dags this is deleted.

RonvanBarneveld (commented):
I have got same commits maybe not really necessary but just curious..

src/dags/airflow_db_permissions.py
```diff
@@ -1,7 +1,7 @@
 import logging
 
 from airflow import DAG
-from common import OTAP_ENVIRONMENT, SLACK_ICON_START, MessageOperator, default_args, slack_webhook_token
+from common import  MessageOperator, default_args
```
ElMudo, line 4: You're going to delete the extra space before MessageOperator at some time, why not now? 😉

src/dags/beheerkaart.py
```diff
@@ -5,12 +5,10 @@
 from bash_env_operator import BashEnvOperator
 from common import (
     OTAP_ENVIRONMENT,
```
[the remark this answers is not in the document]
cvriel: The `OTAP_ENVIRONMENT` is needed to determine the correct folder on the objectstore where to get the sources files. Like 'acceptance' or `production` and such

src/dags/rioolnetwerk.py
cvriel: It uses underwater a temporary named folder for it's temporary data storage. It's included in the SwiftLoadSqlOperator. That operator is doing the download, unzip and ingestion in one go

src/dags/vergunningen.py
cvriel: Same reason as for riool, so basically to get the correct folder in objectstore (acceptance, production and such)

src/dags/airflow_db_permissions.py
cvriel: Ah yes I will, nicely spotted

src/dags/vergunningen.py
```diff
@@ -59,13 +57,9 @@
...
 
-    # 1. Post message on slack
+    # 1. Post info message on slack
     slack_at_start = MessageOperator(
-        task_id="slack_at_start",
-        http_conn_id="slack",
-        webhook_token=slack_webhook_token,
-        message=f"{SLACK_ICON_START} Starting {dag_id} ({OTAP_ENVIRONMENT})",
-        username="admin",
+        task_id="slack_at_start"
```
ElMudo, line 62: Tip: I'd recommend leaving the trailing comma after the last parameter, like you did in the `import` in lines 8-15 and in src/dags/grootstedelijke_projecten.py. Doing so here would reduce this from a six (text) line change to a four line change and make the purpose of the change more evident. This applies to almost all the files here, not just this one. It also has the advantage that if an extra parameter is added later, the change would be two lines shorter than it will in this case.

ElMudo (commented):
As I said in a comment, I'd recommend leaving trailing commas wherever they are tolerated to reduce perceived changes and keep changes cleaner and clearer. The same goes for newly added code.

src/dags/vergunningen.py
cvriel: Yes, I've added the trailing comma after the new added "slack_at_start" parameter. The black layout replaced the import of common on a one liner since some of the imports where omitted due to the new slackOperator

ElMudo (approved):
👍🏻

PR #431, opened 2022-04-07
[UPDATE] roles, users and perms

cvriel:
For the time being we want to give datateams access to their own DAG's in CloudVPS. However since we are running one Airflow instance we need to implement RBAC. This PR includes:

- Creating custom users
- Creating custom roles
- Creating role-bindings
- Adding perms to roles (this is done by the Airflow REST API)

There is no CLI cmd to adding perms to roles. That is why we must use the Airflow REST API to do it dynamically. Alternative is doing it by hand in the Airflow UI. That could be rather cumbersome and error prone. Especially when a user needs to see a lot of DAG's like GOB.

This change is also valid for a relative short period and only applicable for the current situation. When on the new Infra we have multiple instances, an instance for each data team and we do not need to implement RBAC like this.


src/scripts/airflow_high_level_role_perms.sh
```diff
@@ -0,0 +1,38 @@
+# @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ #
+# NOTE: Only needed for CloudVPS. On Azure each datateam has its own Airflow instance.
+#
+# To add generic permissions to the defined roles in `run.sh`, the Airflow API can be used.
+# Unfortuneatly, this cannot be done by the Aiflow CLI. Alternatively you can add the permissions
```
ElMudo, line 5:
```suggestion
# Unfortunately, this cannot be done by the Aiflow CLI. Alternatively you can add the permissions
```

src/scripts/airflow_high_level_role_perms.sh
```diff
@@ -0,0 +1,38 @@
...
+# by hand in the Airflow GUI but that cannot be automated.
+#
+# The specific DAG LEVEL permissions are defined at DAG definition level since some DAG's
```
ElMudo, line 8:
```suggestion
# The specific DAG LEVEL permissions are defined at DAG definition level since some DAGs
```

src/scripts/airflow_high_level_role_perms.sh
```diff
@@ -0,0 +1,38 @@
...
+# like GOB are generated. Listing all DAG's by hand for read and access roles can be
```
ElMudo, line 9:
```suggestion
# such as GOB are generated. Listing all DAGs by hand for read and access roles can be
```

ElMudo (approved):
LGTM. Just some nitpicking remarks on spelling and punctuation in the comments.

src/scripts/airflow_high_level_role_perms.sh
cvriel: :+1:

src/scripts/airflow_high_level_role_perms.sh
cvriel: :+

src/scripts/airflow_high_level_role_perms.sh
cvriel: :+1:

src/Dockerfile
```diff
@@ -79,4 +80,5 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 # Do not use -D to run processes daemonized, does not work!
 # The & puts scheduler in background (not &&)
 # CMD ["sh", "-c", "airflow variables -i vars/vars.json & airflow scheduler & airflow webserver"]
-CMD /run.sh
+# CMD /run.sh
+CMD ["sh", "-c", "/airflow_high_level_role_perms.sh & /run.sh"]
```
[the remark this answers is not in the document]
cvriel: There can only be one CMD line, so I had to combine those two in a one liner. I have keep the original as commented as example how a just a single cmd looks like...

PR #433, opened 2022-04-11
[UPDATE] WIOR

cvriel:
The source endpoint and structure has been altered. This change consists of:

- Defining new data endpoint to retrieve the source data
- Removing some not needed processing of expired attributes within the context of the new data structure


ElMudo (approved):
LGTM (but then again, I'm not familiar with the schema).

PR #460, opened 2022-08-03
AB#51477 Bump flask-caching to v1.11.1

Yrob:
For critical security issue in the old version that showed up in the dependabot alerts.


cvriel:
:+1: thanks for making the changes. We will test this.

cvriel (approved):

PR #519, opened 2023-02-09
Ondergrond: Account for multi geometry types within one table.

cvriel:
Data holds multipolygon and geometry collections. The steps to ingest and to validate the source data must take this duality into account. Currently by using the ingest with ogr2ogr the datatype of the geometry column will be set to multipolygon instead of geometry (generic).


Khelied (approved):
Akkoord

MaguellSandifort (approved):

PR #584, opened 2023-08-09
swap the correct tables for acc

svandenreijen:
Wrong tables were swapped out, fixed reference to the correct tables


RonvanBarneveld (approved):
succes!

